PR #116, opened 2015-02-17
need backslashes for DMD path on Windows

MartinNowak:
- needed for CMD support in dlang.org's win32.mak


MartinNowak:
see https://github.com/MartinNowak/dlang.org/commit/678c57c8ed521fe64a01acab8f456401fb57e26a

PR #117, opened 2015-02-18
fix building dman

MartinNowak:
- DOC and PHOBOSDOC are no longer required


MartinNowak:
> DOC and PHOBOSDOC are no longer required

And actually break the build.

PR #118, opened 2015-02-18
add an option to skip docs building

MartinNowak:


MartinNowak:
We shouldn't delay the beta any longer. So let's add this option and publish v2.067.0-b2 without docs.
Then we'll have another week or so to fix the docs building.

CyberShadow:
> Then we'll have another week or so to fix the docs building.

What else is broken?

PR #119, opened 2015-03-03
build docs only on linux

MartinNowak:
- copy to other platforms
- patch dlang.org to build docs for a release
  from prerelease project layout and disable
  the clone/build targets so that create_dmd_release
  remains the driver


MartinNowak:
The build script collapsed the other day and ceased collaboration with dlang.org's makefile.
In the last minute I noticed that linux 64-bit dmd was actually a 32-bit binary.
Unfortunately this isn't easy to fix as dlang.org insists on building dmd, druntime, and phobos itself.
Another problem with dlang.org is that the docs for a release need to be build like a prerelease.
I ended up patching dlang.org.

The other changes are throwing out plenty of unused code in the create_dmd_release.d script and there are a few smaller fixes.
I'm going to merge this right away in order to build 2.067.0-b3, but I'd be happy for any review.
Also build script experts are welcome to chime in.

PR #121, opened 2015-03-15
Windows installer: Warn before nuking the entire D installation

CyberShadow:
Currently, the uninstaller will nuke the entire D installation directory, including any potential user files or files edited by the user.

Ideally we should only delete the files that we have installed, if they have not been modified.

For now, at least, warn users that we are going to delete everything.

### DIFF
--- a/windows/d1-installer.nsi
+++ b/windows/d1-installer.nsi
@@ -266,8 +266,15 @@ Section "Uninstall"
 
   ; Remove shortcuts
   Delete "$SMPROGRAMS\D\D1 HTML Documentation.lnk"
-  RMDir /r /REBOOTOK "$SMPROGRAMS\D"
+  Delete "$SMPROGRAMS\D\D1 Command Prompt.lnk"
+  RMDir /REBOOTOK "$SMPROGRAMS\D"
 
+  MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION \
+  "The uninstaller will now recursively delete ALL files and directories under '$INSTDIR\dmd'. Continue?" \
+  IDOK rmdir
+  Abort
+
+  rmdir:
   ; Remove used directories
   RMDir /r /REBOOTOK "$INSTDIR\dmd"
   RMDir /REBOOTOK "$INSTDIR"
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -453,8 +453,16 @@ Section "Uninstall"
   ; Remove shortcuts
   Delete "$SMPROGRAMS\D\D2 HTML Documentation.lnk"
   Delete "$SMPROGRAMS\D\D2 Documentation.lnk"
-  RMDir /r "$SMPROGRAMS\D"
+  Delete "$SMPROGRAMS\D\D2 32-bit Command Prompt.lnk"
+  Delete "$SMPROGRAMS\D\D2 64-bit Command Prompt.lnk"
+  RMDir "$SMPROGRAMS\D"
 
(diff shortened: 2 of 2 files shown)


CyberShadow:
This needs testing.

MartinNowak:
I'm not a huge fan of modal yes/no dialogs. They usually only achieve to annoy users, who then click yes/no without reading the message.
You might also have an opinion on a similar topic in dub.
https://github.com/D-Programming-Language/dub/pull/377#issuecomment-80063787

CyberShadow:
Surely you're not saying that silently deleting a directory recursively without a confirmation is better?

I'm aware that this is not a perfect solution, but it is better than nothing.

CyberShadow:
> You might also have an opinion on a similar topic in dub.

We need to be extremely careful about deleting arbitrary files. Most people do not keep backups, and many people are not very organized in where they might keep files. It is entirely imaginable that someone would start editing things in directories that should only be maintained by other programs, because they found those files in the compiler's error messages or for grepping for an identifier.

So, yeah, I would side with Sönke here. Do not delete anything unless you're sure you've created it. I took a similar paranoid approach when I implemented `digger uninstall` - it will refuse to delete anything if the MD5 hashes of files don't match with those that have been installed.

eco:
There is a macro that will [only delete files the installer put there](http://nsis.sourceforge.net/Uninstall_only_installed_files) but it doesn't work with the recursive inclusion we do and wouldn't recognize edited files. To get that we'd need to pregenerate an NSIS include file with a list of files to install as well generate and store a hash of all the files and check those during uninstallation. Kind of a lot of work for a pretty tiny minority of users (though that doesn't preclude doing it since deleting files is serious).

My ideal solution (and one that is a lot more idiomatic on Windows in my opinion) would be to just make it so users never need to edit/add to the installation directory (dmd could find things in `%APPDATA%\D2`). @CyberShadow, what kind of files do you add and change to your installation? Is there any way at all to do what you are doing with an `sc.ini` that includes `%APPDATA%`?

I'm not opposed to this pull request as a stopgap but I but we should get a better solution in place quickly.

CyberShadow:
> My ideal solution (and one that is a lot more idiomatic on Windows in my opinion) would be to just make it so users never need to edit/add to the installation directory (dmd could find things in %APPDATA%\D2).

As per my previous comment, this is futile. We should not be thinking in that direction at all.

> @CyberShadow, what kind of files do you add and change to your installation? Is there any way at all to do what you are doing with an sc.ini that includes %APPDATA%?

This does not concern me personally, but as I've mentioned in the [bug report](https://issues.dlang.org/show_bug.cgi?id=14287), I got complaints from two D users to whom I've recommended D. 

> Kind of a lot of work for a pretty tiny minority of users 

Considering that I don't exactly evangelize D full-time, this does not line up with the theory that only a small % of users are affected by this.

CyberShadow:
> To get that we'd need to pregenerate an NSIS include file with a list of files to install as well generate and store a hash of all the files and check those during uninstallation.

An alternative would be to write a D program which handles actual installation/uninstallation of files (and only use NSIS for the GUI wrapper).

MartinNowak:
Is this established behavior in .msi based installers?
Filed a bug report [18497 – windows uninstaller has an annoying really delete pop-up](https://issues.dlang.org/show_bug.cgi?id=18497).
Relying on custom D tools to install D sounds like an idea with it's own problems. Also adding custom tools to extend the already arcane NSIS sounds questionable from a complexity/debug perspective.

PR #122, opened 2015-04-05
build Win32/COFF library together with phobos64.lib

rainers:
This builds lib32mscoff/phobos32mscoff.lib together with the 64-bit library. It depends on the win64.mak targets added by https://github.com/D-Programming-Language/druntime/pull/1206 and https://github.com/D-Programming-Language/phobos/pull/3150.


andralex:
@WalterBright @MartinNowak okay with this?

CyberShadow:
This is still blocked by the Phobos PR.

MartinNowak:
See #142

PR #125, opened 2015-06-23
update dlang.org patch

MartinNowak:


CyberShadow:
What is this?

MartinNowak:
It's a patch for the dlang.org repo necessary to build documentation for the release.
This is necessary because the prerelease/latest logic doesn't apply.

CyberShadow:
Well, OK. But shouldn't it be integrated in the Makefile instead?

CyberShadow:
I can't review this because I don't know anything about the system, but you have my OK (FWIW) to merge it yourself

MartinNowak:
Ideally we'd add another target to the makefile, but the way it's structured makes it difficult, e.g. it clones the repos again and builds the compiler.

MartinNowak:
It also removes a few menu points that don't apply to the released docs.

PR #128, opened 2015-07-29
Fix issue 14801: OS X installer not compatible with OS X 10.11

jacob-carlborg:
In the upcoming OS X 10.11 /usr is not writable. Instead the compiler
is installed in /Library/D/dmd and symlinked to /usr/local.


MartinNowak:
Thanks a lot Jacob.
It seems that the symlinks make things more complex w/o any benefit, why not use the classical UNIX folders?
If we'd want to use a single dmd folder, I'd prefer /use/local/dmd over /Library/D b/c the library folder contains all the Apple frameworks.

jacob-carlborg:
Short answer: the is only a fix for the installer not working on OS X 10.11, this was easiest to do. /Libraries contains other things as well, Ruby and Perl for example.

MartinNowak:
> /Libraries contains other things as well, Ruby and Perl for example.

Aren't those the versions preinstalled by Apple. Let's go on with this though.
[OS X Mavericks: What is the Library folder?](https://support.apple.com/kb/PH14330?locale=en_US)

MartinNowak:
Should have targeted the stable branch btw.

jacob-carlborg:
> Should have targeted the stable branch btw

Ah, sorry. I'm not familiar with how the stable branches are used in the D projects.

MartinNowak:
Did you actually test this change?
https://issues.dlang.org/show_bug.cgi?id=14801#c12

MartinNowak:
see #133

jacob-carlborg:
Now when you mention it. I never ran the compiler, I only ran the installer and verified that it placed the correct files in the correct locations.

PR #130, opened 2015-08-03
fix Issue 14864 - windows uninstall during installation pops up spurious warning

MartinNowak:
- skip message box when uninstaller is called with /S for silent

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -421,7 +421,7 @@ Function .onInit
     StrCmp $K 0 +2
       Abort
     ; Remove in background the remaining uninstaller program itself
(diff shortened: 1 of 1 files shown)


eco:
I think it'd be better to use `/SD` on the `MessageBox` command to choose a default to go with for silent installs. Like so:

    MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION \
    "The uninstaller will now recursively delete ALL files and directories under '$INSTDIR\dmd2'. Continue?" \
    /SD ID_OK IDOK rmdir

I wonder if there is an NSIS linter...

MartinNowak:
Done, I'll go ahead with this for the RC.

PR #131, opened 2015-08-03
ask to install missing VC/SDK for 64-bit support

MartinNowak:
- detect missing VC/SDK before installation and offer to
  start the VS2013 installation


MartinNowak:
Wait, this shouldn't go into 2.068 yet.

PR #134, opened 2015-08-04
Followup for OS X installer not compatible with 10.11

jacob-carlborg:
* Update dmd2.conf with new install location
* Use sudo for all commands in post install script


MartinNowak:
Let's rename dmd.conf to dmd1.conf to avoid such confusion.

PR #136, opened 2015-08-05
rename dmd2.{conf,pmdoc} -> dmd.{conf,pmdoc}

MartinNowak:


jacob-carlborg:
Should we drop the 2 suffix as well, like DMD2.pkg?

MartinNowak:
Don't want to break any scripts or tools. Renaming the pkg would change the dmg content, right?

jacob-carlborg:
> Renaming the pkg would change the dmg content, right?

Yes.

> Don't want to break any scripts or tools

Ok, then :ok_hand:

PR #138, opened 2015-08-10
fix Issue 14897 - shared linux libraries from zip package don't work

MartinNowak:
[Issue 14897 – shared linux libraries from zip package don't work](https://issues.dlang.org/show_bug.cgi?id=14897)


MartinNowak:
Fairly trivial, this just copies the symlinks for phobos.so into the zip file.

MartinNowak:
If noone cares I'll merge this within the next few days.

PR #139, opened 2015-08-23
keep stdio_msvc12/14.obj files

MartinNowak:


MartinNowak:
No longer needed since https://github.com/D-Programming-Language/druntime/pull/1360.

PR #141, opened 2015-08-25
Move from zip to lzma compressed archive when building Linux packages

jordisayol:


linux/dmd_rpm.sh
```diff
@@ -134,7 +134,7 @@ do
 
 
 		# download zip file if not exist
-		if ! $(unzip -c $DESTDIR"/"$ZIPFILE &>/dev/null)
+		if ! $(tar --lzma -tf $DESTDIR"/"$ZIPFILE &>/dev/null)
```
MartinNowak: `tar -Jtf`?

MartinNowak:
Are these scripts used outside of our build system?

MartinNowak:
It will also be incompatible with #140 b/c the .tar.xz packages are build from the .zip not the other way round. This could prolly be changed though.

jordisayol:
I thought that zip archives will be replaced by "*.tar.xz" and "*.7z" for the next release. My mistake. I close this PR.

MartinNowak:
> I thought that zip archives will be replaced by ".tar.xz" and ".7z" for the next release. My mistake. I close this PR.

We'll have them in addition, maybe deprecate them at some point, but not soon.
It would still be cleaner to generate the zips from the tar.xz, instead of doing it vice versa.

PR #142, opened 2015-08-26
fix Issue 13889 - mscoff32 libs not available

MartinNowak:
[Issue 13889 – mscoff32 libs not available](https://issues.dlang.org/show_bug.cgi?id=13889)


rainers:
LGTM

rainers:
It seems I don't have access rights to merge this...

PR #144, opened 2015-09-03
add support for VS2015 and Universal CRT

rainers:
needs https://github.com/D-Programming-Language/dmd/pull/5031

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -172,6 +172,14 @@ SetCompressor /SOLID lzma
   dandr_done:
 !macroend
 
+; Read SDK registry entry and check if kernel32.lib exists in the expected lib folder
+!macro _DetectSDK REG_KEY VALUE LIBFOLDER
+    ClearErrors
+    ReadRegStr $0 HKLM "Software\Microsoft\${REG_KEY}" ${VALUE}
+    IfErrors +2 0
+    IfFileExists "$0${LIBFOLDER}\kernel32.lib" +2
+    SetErrors
+!macroend
 
 ;--------------------------------------------------------
 ; Interface settings
@@ -273,7 +281,7 @@ SectionGroup /e "D2"
     StrCmp $VCVer "" no_vc_detected write_vc_path
 
     no_vc_detected:
-      MessageBox MB_OK "Could not detect Visual Studio (2008-2013 are supported). No 64-bit support."
+      MessageBox MB_OK "Could not detect Visual Studio (2008-2015 are supported). No 64-bit support."
       goto finish_vc_path
 
     write_vc_path:
@@ -286,7 +294,7 @@ SectionGroup /e "D2"
     StrCmp $WinSDKPath "" no_sdk_detected write_sdk_path
 
     no_sdk_detected:
-      MessageBox MB_OK "Could not detect Windows SDK (6.0A-8.1 are supported). No 64-bit support."
+      MessageBox MB_OK "Could not detect Windows SDK (6.0A-10.0 are supported). No 64-bit support."
       goto finish_sdk_path
 
     write_sdk_path:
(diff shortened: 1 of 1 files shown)


windows/d2-installer.nsi
```diff
@@ -355,10 +367,13 @@ Function DetectVSAndSDK
     StrCpy $VCPath $0
     StrCpy $VCVer $1
 
-    ReadRegStr $0 HKLM "Software\Microsoft\Windows Kits\Installed Roots" "KitsRoot81"
+    ReadRegStr $0 HKLM "Software\Microsoft\Microsoft SDKs\Windows\v10.0" "InstallationFolder"
+    IfErrors 0 done_sdk
+    ClearErrors
+    ReadRegStr $0 HKLM "Software\Microsoft\Microsoft SDKs\Windows\v8.1" "InstallationFolder"
```
MartinNowak: I only have `v8.1a` in my registry and there is no `Lib` folder under `C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1` or `\v8.1a`. Seems like the KitsRoot is correct, at least there is a Lib folder in `C:\Program Files (x86)\Windows Kits\8.1\`.

rainers: I didn't remember why we used these `KitsRoot` entries. For me they are identical to the `Microsoft SDK` entries for the versions without an A. I think we should add a test for the existence of `kernel32.lib` in the expected lib folder.

rainers:
I have added the RootKits registry entries back and added a file check for kernel32.lib.

windows/d2-installer.nsi
```diff
@@ -355,24 +375,49 @@ Function DetectVSAndSDK
     StrCpy $VCPath $0
     StrCpy $VCVer $1
 
-    ReadRegStr $0 HKLM "Software\Microsoft\Windows Kits\Installed Roots" "KitsRoot81"
+	!insertmacro _DetectSDK "Windows Kits\Installed Roots" "KitsRoot10" "lib\um\x64"
```
rainers: I have just guessed the folder to check for the Windows 10 SDK. Is it correct or do we have to use the UCRTVersion as the first folder?

MartinNowak: It's `Lib\10.0.10240.0\um\x64` you have to check.

windows/d2-installer.nsi
```diff
@@ -355,24 +375,49 @@ Function DetectVSAndSDK
...
     IfErrors 0 done_sdk
-    ClearErrors
-    ReadRegStr $0 HKLM "Software\Microsoft\Windows Kits\Installed Roots" "KitsRoot" ; 8.0
+    !insertmacro _DetectSDK "Microsoft SDKs\Windows\v10.0" "InstallationFolder" "lib\um\x64"
```
MartinNowak: There doesn't seem to be anything in those installation folders when installing SDK 8.1, SDK 10.0 or VS2015.

MartinNowak:
We should also update the download url for VS to 2015, http://go.microsoft.com/fwlink/?LinkId=532606.

windows/d2-installer.nsi
rainers: I was just extrapolating from the older versions here. I guess it can be removed.

windows/d2-installer.nsi
rainers: Moved the UCRTVersion detection up to have it available here.

windows/d2-installer.nsi
```diff
@@ -355,26 +375,49 @@ Function DetectVSAndSDK
...
+        StrCmp $1 "" done_ff
+        StrCpy $UCRTVersion $1 ; hoping the directory is retrieved in ascending order (done by NTFS)
+        FindNext $0 $1
+        Goto loop_ff
+      done_ff:
+      FindClose $0
+
+    done:
+
+    !insertmacro _DetectSDK "Windows Kits\Installed Roots" "KitsRoot10" "lib\${UCRTVersion}\um\x64"
```
MartinNowak: It's `Lib` with a capital L.

MartinNowak:
Rest LGTM.

windows/d2-installer.nsi
rainers: Fixed, though it should not make a big difference as file systems on Windows are not case sensitive.

ZombineDev:
Thanks for the good work, guys! I'll test the VS2015 installer support when the next beta comes out.

PR #145, opened 2015-09-05
Fix issue 13234: when updating, uninstaller reports wrong installation directory

rainers:
this bug always makes me nervous because I usually use a changed folder and the uninstaller shows the default one. 
https://issues.dlang.org/show_bug.cgi?id=13234

also reuses old installation path from uninstall folder as default, reported in
https://issues.dlang.org/show_bug.cgi?id=12276

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -377,38 +377,6 @@ Function DetectVSAndSDK
     ClearErrors
 FunctionEnd
 
-; Get path component of file name
- ; Usage:
- ;   Push "C:\Program Files\Directory\Whatever"
- ;   Call GetPath
- ;   Pop $R0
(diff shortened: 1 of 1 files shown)


windows/d2-installer.nsi
```diff
@@ -377,6 +377,38 @@ Function DetectVSAndSDK
     ClearErrors
 FunctionEnd
 
+; Get path component of file name
```
MartinNowak: Use [GetParent](http://sis.sourceforge.net/Docs/AppendixE.html#getparent).

rainers: I just copied the code from the NSIS wiki, I didn't expected to find it in the library, too. Updated.

PR #147, opened 2015-09-19
fix issue - host compiler uses wrong dmd.conf

MartinNowak:
- remove dmd.conf before building dmd from host compiler
- also use dmd.conf target of posix.mak


MartinNowak:
Wrong target branch.

PR #149, opened 2015-09-27
rename link.exe to optlink.exe

MartinNowak:
- still keep a copy as link.exe for now


WalterBright:
I've also added optlink.exe to the Digital Mars zips.

MartinNowak:
> I've also added optlink.exe to the Digital Mars zips.

Much nicer, so we only have a single place to change later on, see #150.

PR #153, opened 2015-10-07
fix for separate changelogs

MartinNowak:
- don't symlink changelog into usr/share/doc/dmd
- it's odd to only have the changelog there
- version number (2.068.2.html) changes with every release


MartinNowak:
@jordisayol

PR #154, opened 2015-10-08
use xz compression for .deb/.rpm packages

MartinNowak:
- supported since many years
- binaries will be much smaller (about 50%)


MartinNowak:
I tested this on a couple of older distributions, the only one that doesn't support xz rpms is CentOS 5, but our rpm package doesn't work for it anyhow.

PR #155, opened 2015-10-13
update default vs version to 2015

MartinNowak:


MartinNowak:
@rainers

ZombineDev:
I think message "No 64-bit support." should be changed to "No 32/64-bit COFF support. The Microsft linker that is included in Visual Studio is required for linking COFF objects" or something like that because, because AFAIU 32mscoff support is the same boat as Win64, that is you need to use Microsoft linker.

rainers:
I'm not 100% sure we should recommend VS2015 as the best experience:
* we are not testing against this version
* Visual D needs some updates, too. I have implemented most of these, but not released them yet. I suspect there are a few more, but I haven't yet switched to VS2015 myself.

MartinNowak:
Your decision, I'm not sure what version runs on the tester.

rainers:
The tester runs VS2012. Let's revisit for the next release.

Geod24:
Time to revisit @rainers ?

PR #157, opened 2015-10-14
create_dmd_release: Copy Win32 import libraries from Druntime

CyberShadow:
For https://github.com/D-Programming-Language/druntime/pull/1402

Untested.

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -467,9 +467,14 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
     if(dmdOnly)
         return;
 
+    string makeTargetDruntime;
+    version(Windows)
+        if (bits == Bits.bits32)
+            makeTargetDruntime = " target implibs";
+
     info("Building Druntime "~bitsDisplay);
     changeDir(cloneDir~"/druntime");
-    run(makecmd~msvcEnv);
+    run(makecmd~msvcEnv~makeTargetDruntime);
     removeFiles(cloneDir~"/druntime", "*{"~obj~"}", SpanMode.depth,
         file => !file.baseName.startsWith("gcstub", "minit"));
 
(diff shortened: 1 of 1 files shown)


MartinNowak:
Could implibs simply become part of the default target?

create_dmd_release/create_dmd_release.d
```diff
@@ -602,6 +607,8 @@ void createRelease(string branch)
         {
             copyFile(cloneDir~"/druntime/lib/gcstub.obj", osDir~"/lib/gcstub.obj");
             copyFile(cloneDir~"/phobos/phobos.lib", osDir~"/lib/phobos.lib");
+            foreach (de; dirEntries(cloneDir~"/druntime/lib/win32/", "*.lib", SpanMode.shallow))
```
MartinNowak: You can use `copyDir(cloneDir~"/druntime/lib/win32/", osDir~"/lib/", file => file.endsWith(".lib"))` instead.

CyberShadow:
> Could implibs simply become part of the default target?

Unfortunately the `implib` utility is not in the autotester's PATH.

MartinNowak:
What's the implib tool? Do I need to install it? That might require to adjust the windows image used to build the release.

CyberShadow:
> What's the implib tool?

Uhh, it creates Windows OMF import libraries from .def files, among other things.

> Do I need to install it?

It's distributed with DMC.

> That might require to adjust the windows image used to build the release.

Does it not have the DMC tools? How does it even build DMD? This doesn't work on the autotester because it uses absolute paths.

MartinNowak:
Let's . How does that work with coff btw, are the files part of VC?

CyberShadow:
Yep. Nothing to do with COFF because unsurprisingly, Microsoft's compilers provide complete import library coverage for Microsoft's OS :)

MartinNowak:
I had to install implib, update the Windows image, and fix win32.mak (https://github.com/D-Programming-Language/druntime/pull/1438) :(.

MartinNowak:
Should be available tomorrow morning with the new nightly.
https://builds.dawg.eu/

PR #161, opened 2015-11-10
fix Issue 15284 - dmd installer hangs when updating installed windows version 

rainers:
there is an infinite loop when no SDK is detected

https://issues.dlang.org/show_bug.cgi?id=15284


MartinNowak:
Thanks, I just starred at the code and suddenly saw it.

PR #162, opened 2015-11-18
installer script

MartinNowak:
- installs requested dmd/gdc/ldc versions
  and dub to a user-wide path (~/dlang) and
  writes sourceable files to activate them
- installs itself to ~/dlang/install.sh
  and can later be used to install/uninstall/list
  compilers or update the script itself
- writes no file outside of the root path, so
  a simple `rm -rf ~/dlang` removes everything
- only requires bash, curl, and xz

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -4,6 +4,7 @@
 # License: Boost License 1.0 (www.boost.org/LICENSE_1_0.txt)
 # Authors: Martin Nowak
 
+_() {
 set -ueo pipefail
 
 # ------------------------------------------------------------------------------
@@ -512,3 +513,6 @@ install_dub() {
 parse_args "$@"
 resolve_latest $compiler
 run_command ${command:-install} $compiler
+}
+
+_ "$@"
--- a/script/install.sh
+++ b/script/install.sh
@@ -66,13 +66,15 @@ check_tools() {
 
 # ------------------------------------------------------------------------------
 
+mkdir -p "$path"
+TMP_ROOT=$(mktemp -d "$path/.installer_tmp_XXXXXX")
+
 mkdtemp() {
-    mkdir -p "$path/.installer_tmp"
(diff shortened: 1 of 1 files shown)


MartinNowak:
`curl -fsSLO https://github.com/MartinNowak/installer/raw/install_script/script/install.sh && bash install.sh dmd && rm install.sh`
The install script will also install itself into ~/dlang/install.sh, so it can be used w/o downloading it again.

eco:
I can't do a real review because I hardly know bash but I did spot check it against a few things from the sandstorm installer (which has had a lot of thought poured into it) and it looks pretty solid.

Getting a certificate for dlang.org so this (and all the other downloads) can be served over HTTPS would be a great first adminstrative act of the D Foundation. I think HTTPS is probably a requirement before this can be linked to from the download page.

MartinNowak:
Yes HTTPS would be great, though we want the downloads to be cacheable. I still want to add gpg verification to the script.

MartinNowak:
I use the trick to wrap the whole script in a function from the sandstorm installer, https://github.com/MartinNowak/installer/commit/aa15ed84b126febb9389d8df167b1ac8934f299a and am now verifying gpg signatures https://github.com/MartinNowak/installer/commit/ce2eca5e3cb887e0122062cdae20e61d318333e8.

MartinNowak:
Alternative download from `curl -fsSL https://builds.dawg.eu/install.sh | bash -s dmd`.

ZombineDev:
Martin, this script is already used for Travis CI (and maybe some other places), making it somewhat official. It makes no sense to not be included in the official installer repo. Please merge this PR, so people that want to use it (like myself) can rely on getting it from here.
Btw, thanks for the good work :)

MartinNowak:
Fixed a few final things and u.ploaded (for preview) to dlang.org.

PR #164, opened 2016-01-16
Bump version of Visual D to 0.3.43

rainers:


MartinNowak:
Sorry, only saw it now after building 2.070.0-b2.

PR #165, opened 2016-01-17
modify access rights for sc.ini to allow all users to read the file

rainers:
The permission is lost during modification by _ReplaceInFile.

You'll need to install the AccessControl-plugin for NSIS to build this.

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -309,6 +309,10 @@ SectionGroup /e "D2"
     no_ucrt_detected:
       ClearErrors
 
+    ; Make sc.ini read accessible by all users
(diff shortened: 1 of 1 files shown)


eco:
Didn't see you did this. I have an alternative version: https://github.com/D-Programming-Language/installer/pull/166

Not sure which is better.

CyberShadow:
Both PRs seem to set a specific set of permissions. Ideally we'd want to **delete** all file-specific permissions, and make the file inherit the directory's permissions - meaning that if DMD has been installed to a user-writable directory, the .ini file should be writable too, but not if it was installed e.g. under Program Files.

rainers:
It's a bit annoying to have the dependency on a non-standard NSIS plugin, so I'll close this in favor of PR #166.

PR #166, opened 2016-01-18
Fix sc.ini permissions issue on Windows 10

eco:


CyberShadow:
If you upload an installer EXE, I can test it in an XP VM if you like.

eco:
http://gnuk.net/dmd-2.069.2.exe

I might change this to just have it inherit the permissions like you mentioned in the other pull request.

CyberShadow:
The installation seemed to have been completed successfully.

eco:
A silent failure is pretty much exactly what we wanted to happen.

I've pushed a new version that follows you recommendation to inherit permissions instead of granting read to Users. I just did it on the whole installation directory to catch everything.

rainers:
The problem is not specific to Windows 10, I can reproduce it with 8.1, and it should exist for all versions that allow restricted users including XP IIRC. `icacls` seems available starting with Windows 8, so it might not fit all cases. I think we should still support Windows 7, maybe by a fallback to `cacls`.

CyberShadow:
@rainers Is there an AccessControl API to reset ACLs on a file? If so then we should use that.

CyberShadow:
BTW, there should be a Bugzilla issue for this, linking to the forum thread and these discussions, and a link to the Bugzilla in a code comment.

CyberShadow:
Would it be realistic to fix the root of the issue (the string replace function using a system temp directory) instead?

CyberShadow:
Alternatively, does NSIS provide a way to copy a file's contents but not attributes?

rainers:
Bugzilla entries are https://issues.dlang.org/show_bug.cgi?id=15456 and https://issues.dlang.org/show_bug.cgi?id=15572

rainers:
> Would it be realistic to fix the root of the issue (the string replace function using a system temp directory) instead?

Seems easy: just replace
```
  GetTempFileName $R2               ; who's new?
```
with
```
  Push "$2.tmp"                     ; temp file in same directory
  Pop $R2
```
in ReplaceInFile.nsh

rainers:
I also noticed that the file owner of sc.ini could not be displayed. Not going through the temp folder fixed that, too.
I've made a new PR with the change to ReplaceInFile.nsh: https://github.com/D-Programming-Language/installer/pull/167

WalterBright:
> I think we should still support Windows 7

This exact problem was reported to me by a 64 Windows 7 Pro user.

In any case, we must still support Windows 7. A lot of developers are not wanting to move to newer versions of Windows.

MartinNowak:
My understanding is that #167 supercedes this PR, so can it be closed?

eco:
icacls is available in Windows 7. It has been included in every version of Windows since 2003. #167 works for me anyway, though.

eco:
Ugh, sorry. I made a commit 63b9dc5015 accidentally (instead of a Pull Request). Normally when you hit Edit on Github it will make a pull request but apparently since I'm a committer on this repo it just went ahead and made a commit.

PR #168, opened 2016-01-18
update dlang.org patch for release docs

MartinNowak:


MartinNowak:
Needed to fix the nightly build.

PR #180, opened 2016-06-17
update dlang.org patch

MartinNowak:


CyberShadow:
Not to stick my nose where my pot isn't boiling, but this seems like a nontrivial maintenance effort. Why not put this in the dlang.org repo, enabled by adding a .ddoc file?

PR #181, opened 2016-07-06
build dub with dmd releases

MartinNowak:
- dub prereleases are only use for dmd
  prereleases, not for nightlies and dmd releases

@s-ludwig


andralex:
Cool, thanks! I think we need to bring more collaborators to this repo so you can get your code reviewed properly.

MartinNowak:
Yes that would helpful, but create_dmd_release.d, the actual build script that runs on each platform is still quite a mess.

PR #184, opened 2016-08-01
Merge remote-tracking branch 'upstream/stable' into merge_stable

MartinNowak:


MartinNowak:
Only contains cherry-picks which result in lots of merge conflicts but no actual changes.

PR #185, opened 2016-08-02
move xdg-utils from Depends to Suggests

MartinNowak:
- fixes the problem that dmd currently installs a complete X11 server

- only needed for browser/desktop related functionality, e.g. dman, or
  std.phobos.brower open, it's OK when those don't work on a headless server


jordisayol:
If you think that this is better, for me no problem.

mathias-lang-sociomantic:
Thanks A LOT !

PR #187, opened 2016-08-06
fix Issue 16349 - better curl retry for install.sh script

MartinNowak:
- curl's --retry option isn't really helpful b/c no timeouts are
  activated by default
- use a retry loop with increasing sleep times instead
- enable connection (5s) and download (<1KB/s for 30s) timeouts
  (using --max-time is too tricky b/c of the unknown download sizes)

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright: Copyright 2015 Martin Nowak.
+# Copyright: Copyright Martin Nowak 2015 -.
 # License: Boost License 1.0 (www.boost.org/LICENSE_1_0.txt)
 # Authors: Martin Nowak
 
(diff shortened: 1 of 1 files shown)


MartinNowak:
FYI @wilzbach

script/install.sh
```diff
@@ -34,15 +34,47 @@ fatal() {
...
+    local url=$1
+    local path=$2
+
+    for i in {0..4}; do
+        if curl "$url" -o "$path"; then
+            break
+        elif [ $i -lt 4 ]; then
+            sleep $((1 << $i))
+        else
+            fatal "Failed to download '$url'"
```
wilzbach: Should we try to cleanup intermediate temp files on failure?
(due to `fatal` exiting directly the cleanup in `fetch` won't take place)

MartinNowak: There is a cleanup trap that wipes the whole TMP_ROOT.

MartinNowak:
I tested all compilers, and the `--activate` functionality. Also played a bit w/ tc to simulate a bad network connection, but it's tricky to properly test this. It should be better than before.

wilzbach:
> I tested all compilers

Maybe at some point we should setup a small travis CI for this repo that does this automatically? ;-)

> Also played a bit w/ tc to simulate a bad network connection, but it's tricky to properly test this. It should be better than before.

I did the same and also tried it a bit and it looks really good - e.g. I interrupted the network during the download and all the retry and timeouts fire correctly :)

```
Downloading and unpacking http://downloads.dlang.org/releases/2.x/2.071.1/dmd.2.071.1.linux.tar.xz
##                                                                         3.4%



curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds

curl: (7) Couldn't connect to server

curl: (7) Couldn't connect to server

curl: (7) Couldn't connect to server

curl: (7) Couldn't connect to server
Failed to download 'http://downloads.dlang.org/releases/2.x/2.071.1/dmd.2.071.1.linux.tar.xz'
```

So from my side this looks great - thanks for doing this @MartinNowak :)

MartinNowak:
Let's closely monitor if that fixes the remaining Travis-CI issues.

PR #188, opened 2016-08-06
fix unbound bash variables in activate script

MartinNowak:
- allow to source the activate script with set -u


MartinNowak:
Just providing empty defaults.

PR #189, opened 2016-08-08
Download DMD by default

wilzbach:
Small follow-up on #1437 - I assume if you (@MartinNowak) thought that downloading DMD would be the default action, it's a good justification to make it so. On the other hand showing a help page if no command is provided is a good style, but I assume that this script is mostly only executed directly from the web or in CI scripts (and `-h` | `--help` still exists).

Btw [Rust](https://www.rust-lang.org/en-US/downloads.html) also downloads the compiler by default.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -78,7 +78,7 @@ fetch() {
 # ------------------------------------------------------------------------------
 
 command=
-compiler=
+compiler=dmd
 verbosity=1
 path=~/dlang
 case $(uname -s) in
@@ -147,6 +147,7 @@ Options
   -v --verbose  Verbose output
 
 Run "install.sh <command> --help to get help for a specific command.
+If no argument are provided, the latest DMD compiler will be installed.
 '
 }
 
@@ -154,10 +155,10 @@ command_help() {
     local _compiler='Compiler
 
   dmd|gdc|ldc           latest version of a compiler
-  dmd|gdc|ldc-<version> specific version of a compiler (e.g. dmd-2.069.0, ldc-0.16.1-beta2)
+  dmd|gdc|ldc-<version> specific version of a compiler (e.g. dmd-2.071.1, ldc-1.1.0-beta2)
   dmd-beta              latest dmd beta
   dmd-nightly           latest dmd nightly
-  dmd-2015-11-22        specific dmd nightly
+  dmd-2016-08-08        specific dmd nightly
 '
 
     case $1 in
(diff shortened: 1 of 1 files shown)


script/install.sh
```diff
@@ -231,6 +231,11 @@ parse_args() {
     local _help=
     local _activate=
 
+    # select DMD by default
+    if [[ $# -eq 0 ]] ; then
+        compiler="dmd"
+    fi
```
MartinNowak: Let's rather set the default value [here](https://github.com/wilzbach/installer/blob/af19e74f72ae938c353337d6be3b367a738e6be8/script/install.sh#L81).

wilzbach:
> Let's rather set the default value here.

That's a lot better. Thanks!

MartinNowak:
Let's also update the help https://github.com/wilzbach/installer/blob/af19e74f72ae938c353337d6be3b367a738e6be8/script/install.sh#L179 with an example and mentioning that dmd is the default.

script/install.sh
```diff
@@ -78,7 +78,7 @@ fetch() {
 # ------------------------------------------------------------------------------
 
 command=
```
MartinNowak: While we're at it, would be great to replace the other default argument mechanism `${command:-install}` with setting command here.

wilzbach: This doesn't work as then `install` will always be selected as help page.

script/install.sh
```diff
@@ -169,17 +170,19 @@ command_help() {
 Description
 
   Download and install a D compiler.
+  By default the latest release of the DMD compiler is selected.
```
wilzbach: I wasn't sure whether it's better to have this text here or besides the examples below.

script/install.sh
```diff
@@ -147,17 +147,18 @@ Options
...
+If no argument are provided, the latest DMD compiler will be installed.
 '
 }
 
 command_help() {
     local _compiler='Compiler
 
   dmd|gdc|ldc           latest version of a compiler
-  dmd|gdc|ldc-<version> specific version of a compiler (e.g. dmd-2.069.0, ldc-0.16.1-beta2)
+  dmd|gdc|ldc-<version> specific version of a compiler (e.g. dmd-2.071.1, ldc-1.1.0-beta2)
```
wilzbach: I thought it's nice if the examples are a bit up-to-date, but I can revert that if you see it as noise.

script/install.sh
MartinNowak: Ah, now I remember.

PR #190, opened 2016-08-14
Update libcurl-7.50.1

brocolis:
http://d.darktech.org/libcurl-7.50.1-WinSSL-zlib-x86-x64.zip
MD5: 028280a8a80254ec420a32db2fc73220

Please upload it to http://downloads.dlang.org/other/


andralex:
I'll pull. @MartinNowak please lmk if I did something silly,

andralex:
Actually I better defer this to @MartinNowak ...

PR #191, opened 2016-08-15
remove multiarch-support from Pre-Depends field in phobos deb package

jordisayol:


MartinNowak:
Why? It's always helpful to provide context for PRs.

PR #196, opened 2016-09-23
retry fetch to stdout operations

MartinNowak:
- might fails as much as downloads


MartinNowak:
Not sure why I "forgot" this method in #187, hoping for shorter downloads to be more reliable was naive.

PR #198, opened 2016-09-28
support building previews for feature branches

MartinNowak:
- fallback to master if a feature branch does not
  exist in some of the dlang repos

@UplinkCoder


MartinNowak:
Just going to merge this, as I'm still more or less the only maintainer of the build scripts.

PR #201, opened 2016-10-21
Fix installation of dmd point releases

wilzbach:
Seems like #200 broke the ability to download/install point releases as it tries to query for the latest build version. This just extends the ignore regex to add point releases

CC @MartinNowak

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -368,8 +368,12 @@ resolve_latest() {
             compiler="dmd-$(fetch $url)"
             ;;
         dmd-*) # nightly master or feature branch
-            # dmd-nightly, dmd-master, dmd-branch (but not dmd-2016-10-19 or dmd-branch-2016-10-20)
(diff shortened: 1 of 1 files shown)


andralex:
I'll pull this (for) now. Thanks.

wilzbach:
> I'll pull this (for) now. Thanks.

It seems that the file at https://dlang.org/install.sh is still using the old version. Does anyone know how we can update it?

MartinNowak:
Thanks and sorry for the troubles, time to write tests for that script ;).

MartinNowak:
The script is simply copied to dlang.org, i.e. `scp install.sh install.sh.sig martin@digitalmars.com:/var/pub/digitalmars.com/`. We have a couple of people who can do that.
It's currently done manually b/c I'm also signing the script. Would be good to have more people in our keyring http://dlang.org/gpg_keys.html.

PR #211, opened 2017-02-06
[WIP] use SSL for downloads

wilzbach:
https://trello.com/c/mWHZxwo5/89-https-for-s3-downloads-dlang-org

At the moment this doesn't work because it seems that there's no website redirection in place:

```
>  curl -v https://s3-us-west-2.amazonaws.com/downloads.dlang.org/releases/2.x/2.073.0/dmd.2.073.0.linux.tar.xz
...
< x-amz-website-redirect-location: http://downloads.dlang.org/releases/2017/dmd.2.073.0.linux.tar.xz
< Content-Length: 0
```

The problem is that the current folder structure is year-based and depends on dummy objects with the `x-amz-website-redirect-location` for redirects. However AWS S3 handles [page redirects]http://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html) differently:

Region-specific website endpoint – Amazon S3 redirects the page request according to the value of the x-amz-website-redirect-location property.
REST endpoint – Amazon S3 does not redirect the page request. It returns the requested object.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -7,6 +7,8 @@
 _() {
 set -ueo pipefail
 
+dlang_download_mirror=https://s3-us-west-2.amazonaws.com/downloads.dlang.org
+
 # ------------------------------------------------------------------------------
 
 log_() {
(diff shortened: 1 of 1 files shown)


wilzbach:
As there doesn't seem to be an easy solution and the binaries are signed, closing for now.

PR #214, opened 2017-02-26
Check whether DUB is provided by the release

wilzbach:
DUB is provided by DMD (>= 2.072.0) and LDC (>= 1.1.0), so there's no need to download it again.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -295,6 +295,7 @@ parse_args() {
 run_command() {
     case $1 in
         install)
+            local has_dub_included=0
             check_tools curl
             if [ ! -f "$path/install.sh" ]; then
                 install_dlang_installer
@@ -306,8 +307,11 @@ run_command() {
                 log "$2 already installed";
             else
                 install_compiler $2
+                has_dub_included=$?
             fi
-            install_dub
+            if [ "$has_dub_included" -eq 0 ] ; then
+                install_dub
+            fi
             write_env_vars $2
 
             if [ $(basename $SHELL) = fish ]; then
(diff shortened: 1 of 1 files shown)


MartinNowak:
Will merge and release later today.

script/install.sh
```diff
@@ -417,6 +423,10 @@ install_compiler() {
             local arch="zip"
         fi
 
+        if [[ $ver > "2.072.0z" ]]; then
```
wilzbach: Hmm on a second though: this seems to exclude `2.072.0`, should I change it to sth. like `2.071.9`?

script/install.sh
```diff
@@ -443,6 +453,9 @@ install_compiler() {
         if [ $os != linux ] && [ $os != osx ]; then
             fatal "no ldc binaries available for $os"
         fi
+        if [[ $ver > "1.1.0z" ]]; then
```
wilzbach: similarl here: should I change it to `1.0.99`?

MartinNowak: That's what the `z` is for `1.0z` should work.

script/install.sh
MartinNowak: `2.071z`

MartinNowak:
Had to enable squash merges for this repo first.

script/install.sh
```diff
@@ -306,8 +307,11 @@ run_command() {
                 log "$2 already installed";
             else
                 install_compiler $2
+                has_dub_included=$?
+            fi
+            if [ "$has_dub_included" -eq 0 ] ; then
+                install_dub
             fi
```
MartinNowak: Mmh, this still queries code.dlang.org when a compiler is already installed.

MartinNowak:
You also didn't adopt the activate scripts to work w/o a separate dub installation.

wilzbach:
> You also didn't adopt the activate scripts to work w/o a separate dub installation.

Hmm, it still "works", because it's sets the `$PATH` variable.
However, I agree it would be nicer not to put the DUB path in there, if it doesn't exist (-> will do a short follow-up).

MartinNowak:
Also seems to fail, at least during initial installation, due to `return 1` and `set -e` (errexit).

MartinNowak:
I'll revert this for now to not impact any usage.

PR #215, opened 2017-02-28
update dlang.org patch

MartinNowak:


MartinNowak:
Merging now in order to rebuild nightly.

wilzbach:
Sorry - I didn't know that you were patching the Makefile. What changes do you need?
Maybe we can make this happen via variables (or targets)?

MartinNowak:
Had that discussion a couple of times, the release builds a hybrid documentation where the `*-release` targets use pre-release dmd/druntime/phobos folders, b/c obviously the latest doesn't match.
It also disables clone and build triggers, so the Makefile doesn't interfere w/ the build script.

There are also some minor tweaks to remove download and pre-release links.

wilzbach:
> Had that discussion a couple of times

It's my first time :)

> the release builds a hybrid documentation where the *-release targets use pre-release dmd/druntime/phobos folders, b/c obviously the latest doesn't match.

those are now variables - we can make them optional, s.t. you can directly overwrite them

```
DMD_STABLE_DIR=${DMD_DIR}-${LATEST}
DRUNTIME_STABLE_DIR=${DRUNTIME_DIR}-${LATEST}
PHOBOS_STABLE_DIR=${PHOBOS_DIR}-${LATEST}
```

> It also disables clone and build triggers, so the Makefile doesn't interfere w/ the build script.

Could be a `NO_BUILD` variable ...

wilzbach:
> @MartinNowak added the auto-merge label 34 minutes ago

Btw Travis seems to be pretty clogged at the moment. This might take a couple of hours :/

PR #217, opened 2017-03-02
partly fix docs building

MartinNowak:


MartinNowak:
Still having more issues, e.g. building with LATEST=master no longer works (and maybe rightful so).
I'll disable docs for the nightlies until all of those dlang.org changes are done and I find time to fix the fallout.

wilzbach:
> e.g. building with LATEST=master no longer works (and maybe rightful so).

How did this ever work?
I didn't touch this target:

```
../%-${LATEST} :
    git clone -b v${LATEST} --depth=1 ${GIT_HOME}/$* $@
```

As `vmaster` doesn't exist, it fails:

```
git clone -b vmaster --depth=1 https://github.com/dlang/druntime ../druntime-master
Cloning into '../druntime-master'...
warning: Could not find remote branch vmaster to clone.
fatal: Remote branch vmaster not found in upstream origin
make: *** [posix.mak:350: ../druntime-master] Error 128
```

> Still having more issues

I am really sorry about this and would love to help.
Please send me then a list of everything that you would like to be configurable, s.t. we can avoid the error-prone patching here.

PR #218, opened 2017-03-02
Check for dub

wilzbach:
> Also seems to fail, at least during initial installation, due to return 1 and set -e (errexit).

Can't reproduce this. Added tests for it to let Travis verify this.

> Mmh, this still queries code.dlang.org when a compiler is already installed.

Added a check. My main motivation was really just to avoid unnecessary downloads on Travis 

>> You also didn't adopt the activate scripts to work w/o a separate dub installation.
> It still "worked", because of the $PATH variable.

Now the `dub` folders shouldn't been written to the `activate` scripts anymore.

> I'll revert this for now to not impact any usage.

Sorry about that :/

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -91,6 +91,7 @@ COMMAND=
 COMPILER=dmd
 VERBOSITY=1
 ROOT=~/dlang
+DUB_BIN_PATH=
 case $(uname -s) in
     Darwin) OS=osx;;
     Linux) OS=linux;;
@@ -323,7 +324,16 @@ run_command() {
             else
                 install_compiler "$2"
             fi
-            install_dub
+
+            local -r dub_bin_path=$(find_dub "$2")
+            if [ -f "${dub_bin_path}/dub" ] ; then
+                local -r dub_version=$("${dub_bin_path}/dub" --version | cut -f3 -d' ' | tr -d ,)
+                log "dub ${dub_version} already installed";
+            else
+                DUB_BIN_PATH="${ROOT}/dub"
+                install_dub
+            fi
+
             write_env_vars "$2"
 
             if [ "$(basename "$SHELL")" = fish ]; then
@@ -488,6 +498,26 @@ install_compiler() {
     fi
 }
 
+# Checks whether a DUB executable is already included in the release
+find_dub() {
+    case $1 in
+        dmd*)
+            local suffix=
+            if [ "${OS:-}" != "osx" ] ; then
+                local suffix=$MODEL
+            fi
+            local bin_path="${OS}/bin${suffix}"
+            ;;
+        ldc*)
+            local bin_path=bin
+            ;;
+        *)
+            local bin_path=
+            ;;
+    esac
+    echo "$ROOT/$1/$bin_path"
+}
+
 find_gpg() {
     if command -v gpg2 &>/dev/null; then
         echo gpg2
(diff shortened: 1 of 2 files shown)


wilzbach:
@MartinNowak: can we move forward with this?
Downloading dub from code.dlang.org adds yet another server which is rather unstable to the mix:

https://travis-ci.org/dlang/dmd/jobs/276138166

(especially when it's already existent)

MetaLang:
Doesn't this mean that dub could get out of date on the host machine?

wilzbach:
> Doesn't this mean that dub could get out of date on the host machine?

You meant that you have an old D compiler and still want to use the newest DUB? Well, definitely a use case, but downloading a newer DUB yourself isn't that difficult ...
However, note that there's also the contrary point that newer DUB versions might drop (some) support for old compilers.

The point of this issue / PR was that on the CI machines we essentially download DUB twice and in particular the download from code.dlang.org is problematic as code.dlang.org has longer downtimes fairly regularly.

MartinNowak:
We could add explicit installs for e.g. `dub-1.6.0`, would need a separate, nestable activate/deactivate script though.

script/install.sh
```diff
@@ -320,10 +321,35 @@ run_command() {
...
+                    log "dub ${dub_version} already installed";
+                fi
             else
                 install_compiler "$2"
             fi
-            install_dub
+
+            if [ "$HAS_DUB_INCLUDED" -eq 0 ] ; then
+                install_dub
+            fi
```
MartinNowak: Why do you use a version based approach below, and a path based approach here?

wilzbach:
Okay I finally found time to rebase this again and reduce its size. Now there's no version checking anymore and it solely checks whether the DMD/LDC release comes with a DUB binary.

script/install.sh
```diff
@@ -592,9 +622,9 @@ _OLD_D_LIBRARY_PATH="\${LIBRARY_PATH:-}"
 _OLD_D_LD_LIBRARY_PATH="\${LD_LIBRARY_PATH:-}"
 _OLD_D_PS1="\${PS1:-}"
 
-export PATH="$ROOT/dub:$ROOT/$1/$binpath:\${PATH:-}"
-export LIBRARY_PATH="$ROOT/$1/$libpath:\${LIBRARY_PATH:-}"
-export LD_LIBRARY_PATH="$ROOT/$1/$libpath:\${LD_LIBRARY_PATH:-}"
+export PATH="${DUB_BIN_PATH}${DUB_BIN_PATH:+:}$ROOT/$1/$binpath\${PATH:+:}\${PATH:-}"
+export LIBRARY_PATH="$ROOT/$1/$libpath\${LIBRARY_PATH:+:}\${LIBRARY_PATH:-}"
+export LD_LIBRARY_PATH="$ROOT/$1/$libpath\${LD_LIBRARY_PATH:+:}\${LD_LIBRARY_PATH:-}"
```
MartinNowak: A trailing `:` means that ld.so would load shared libraries from the current working directory, thus hiding system wide shared libraries :o.

MartinNowak:
Changed a few things, could you have a look @wilzbach.

wilzbach:
>  Changed a few things, could you have a look @wilzbach.

CI was failing with:

```
./script/install.sh: line 634: bad substitution: no closing `}' in '}${DUB_BIN_PATH}${DUB_BIN_PATH:+' 
```

-> I escaped the parameter :wink: 

> A trailing : means that ld.so would load shared libraries from the current working directory, thus hiding system wide shared libraries :o.

Wow good catch. I like the `:+:` parameter expansion trick. Nice one!

PR #227, opened 2017-04-21
Windows: add detection of VS 2017

rainers:
Needs https://github.com/dlang/dmd/pull/6716

See https://issues.dlang.org/show_bug.cgi?id=17280 and https://issues.dlang.org/show_bug.cgi?id=17320

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -94,6 +94,7 @@
 !include "EnvVarUpdate.nsh"
 !include "ReplaceInFile.nsh"
 !include "FileFunc.nsh"
+!include "TextFunc.nsh"
 
 
 ;------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rainers:
Should we be more conservative and prefer one of the older versions if multiple VS are installed?

WalterBright:
Prefer the latest.

windows/d2-installer.nsi
```diff
@@ -351,6 +352,15 @@ SectionGroupEnd
 Function DetectVSAndSDK
     ClearErrors
 
+    ReadRegStr $0 HKLM "SOFTWARE\Microsoft\VisualStudio\SxS\VS7" "15.0"
+    StrCpy $1 "VC2017"
+    IfErrors not_vc2017
+        ${LineRead} "$0\VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt" "1" $2
```
MartinNowak: For reference [NSIS 2.46 Users Manual#LineRead](http://anders_k.users.sourceforge.net/public/NSIS/2.46/Manual.html#lineread).

windows/d2-installer.nsi
```diff
@@ -351,6 +352,15 @@ SectionGroupEnd
 Function DetectVSAndSDK
     ClearErrors
 
+    ReadRegStr $0 HKLM "SOFTWARE\Microsoft\VisualStudio\SxS\VS7" "15.0"
```
kdmult: I have installed VS 2017 Pro and there is no key HKLM\SOFTWARE\Microsoft\VisualStudio\SxS\VS7.
But there is HKLM\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\SxS\VS7.

rainers:
> But there is HKLM\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\SxS\VS7.

The installer is a 32-bit process so the key is translated to the Wow6432Node by a 64-bit OS.

wilzbach:
This should have been targeted at stable. I cherry-picked it, s.t. it's part of the 2.074.2 release ;-)

https://github.com/dlang/installer/pull/228

windows/d2-installer.nsi
MartinNowak: See answer below https://github.com/dlang/installer/pull/227#issuecomment-299717799.

PR #230, opened 2017-06-07
update dlang.org patch

MartinNowak:
- for added dmd docs


wilzbach:
I think this should error-prone patching needs to stop :cry: 
Maybe https://github.com/dlang/dlang.org/issues/1681 helps?

PR #239, opened 2017-07-17
install.sh: Fix ShellCheck warnings

CyberShadow:


script/install.sh
```diff
@@ -619,13 +627,13 @@ uninstall_compiler() {
         fatal "$1 is not installed in $path"
     fi
     log "Removing $path/$1"
-    rm -rf "$path/$1"
+    rm -rf "${path:-.}/$1"
```
CyberShadow: This one is particularly scary - it doesn't look like it can expand to `rm -rf /`, but it gets too close for comfort, e.g. with `--path /`.

script/install.sh
```diff
@@ -619,13 +627,13 @@ uninstall_compiler() {
...
 }
 
 list_compilers() {
     check_tools egrep
     if [ -d "$path" ]; then
-        ls "$path" | egrep -v '^dub|^install\.sh|^d-keyring\.gpg'
+        find "$path" -maxdepth 1 -printf "%f\n" 1 | grep -Ev '^dub|^install\.sh|^d-keyring\.gpg'
```
ZombineDev: Why the change from `ls` to `find -maxdepth 1`?

script/install.sh
ZombineDev: Indeed.

CyberShadow:
> It would have been nice if each category of warnings fixed was in its own commit and the commit message included the warning text itself.

Good idea, done.

script/install.sh
CyberShadow: As I understand, `ls`'s output is not meant to be parseable, it's meant for humans. The particular warning here is [SC2012](https://github.com/koalaman/shellcheck/wiki/SC2012), but there are some other related warnings related to piping `ls`.

Note that this fix was incorrect and suboptimal; in the latest version it is all one `find` invocation, which also allows us to drop `grep`/`egrep`.

wilzbach:
Btw how about adding a check to Travis, s.t. it future additions are checked as well?

-> https://github.com/koalaman/shellcheck/wiki/TravisCI

script/install.sh
```diff
@@ -619,7 +619,7 @@ uninstall_compiler() {
         fatal "$1 is not installed in $path"
     fi
     log "Removing $path/$1"
-    rm -rf "$path/$1"
+    rm -rf "${path:?}/$1"
```
MartinNowak: Well that's why we're using `set -u`, but anyhow.

script/install.sh
```diff
@@ -72,8 +72,9 @@ download() {
 
 # url
 fetch() {
-    local url=$1
-    local path=$(mktemp "$TMP_ROOT/XXXXXX")
+    local url path
+    url=$1
+    path=$(mktemp "$TMP_ROOT/XXXXXX")
```
MartinNowak: neat

script/install.sh
CyberShadow: With `set -u`, `$var` will error only if it's unset, not empty. `${var:?}` will always error if `var` is empty (or unset).

PR #240, opened 2017-08-08
disable dman to workaround Issue 17731

MartinNowak:


MartinNowak:
Merging since it's blocking 2.075.1-b1

PR #244, opened 2017-08-14
update dlang.org patch

MartinNowak:


create_dmd_release/patches/dlang.org.patch
```diff
@@ -25,22 +25,23 @@ index b18204f..745fc99 100644
...
+@@ -126,6 +122,15 @@ DDOC_VARS_STABLE_HTML=\
  	  DOCSRC="$(PWD)" \
  	  VERSION="$(abspath ${DMD_DIR}/VERSION)"
  
-+DDOC_VARS_RELEASE=\
++DDOC_VARS_RELEASE_HTML=\
 +	  DOC_OUTPUT_DIR="${DOC_OUTPUT_DIR}/phobos" \
 +	  STDDOC="$(addprefix $(PWD)/, $(STD_DDOC))" \
 +	  DMD="$(abspath $(DMD))" \
++	  DMD_DIR="$(abspath ${DMD_DIR})" \
```
MartinNowak: DMD_DIR is needed since we include dmd/osmodel.mak from the recursive make invocation in .generated/phobos.

create_dmd_release/patches/dlang.org.patch
```diff
@@ -84,50 +85,32 @@ index b18204f..745fc99 100644
...
-     $(DIVC head,
-         $(H2 Library Reference)
--        $(P $(SPANC smallprint, version $(LATEST) $(SPANC separator, $(BR))
--            switch to
--            $(LINK2 ../phobos-prerelease/index.html, pre-release). $(BR)
--            $(LINK2 ../library/index.html, preview) beta library reference.
--            )
--        )
-+        $(P $(SPANC smallprint, version $(LATEST)))
-         $(P $(LINK2 index.html, overview))
```
MartinNowak: No longer needed with the new version-changer widget (which is JS and shouldn't run in chm).

PR #251, opened 2017-09-04
Update gdc download path

Bobo1239:
example version: gdc6.3.0+2.068.2 -> http://gdcproject.org/downloads/binaries/6.3.0/x86_64-linux-gnu/gdc-6.3.0+2.068.2.tar.xz

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -460,7 +460,7 @@ install_compiler() {
             x86_64) local triplet=x86_64-linux-gnu;;
             x86) local triplet=i686-linux-gnu;;
         esac
-        if [[ $ver > "4.8.2" ]]; then
+        if [[ $ver > "5.2.0" ]]; then
             local url="http://gdcproject.org/downloads/binaries/$ver/$triplet/$name.tar.xz"
         else
             local url="http://gdcproject.org/downloads/binaries/$triplet/$name.tar.xz"
--- a/script/install.sh
+++ b/script/install.sh
@@ -460,7 +460,11 @@ install_compiler() {
             x86_64) local triplet=x86_64-linux-gnu;;
             x86) local triplet=i686-linux-gnu;;
         esac
-        local url="http://gdcproject.org/downloads/binaries/$ver/$triplet/$name.tar.xz"
+        if [[ $ver > "4.8.2" ]]; then
+            local url="http://gdcproject.org/downloads/binaries/$ver/$triplet/$name.tar.xz"
+        else
+            local url="http://gdcproject.org/downloads/binaries/$triplet/$name.tar.xz"
+        fi
 
         download_and_unpack "$url" "$path/$1"
 
--- a/script/install.sh
+++ b/script/install.sh
@@ -450,8 +450,9 @@ install_compiler() {
         download_and_unpack "$url" "$path/$1"
 
     # gdc-4.8.2, gdc-4.9.0-alpha1, gdc-5.2, or gdc-5.2-alpha1
-    elif [[ $1 =~ ^gdc-([0-9]+\.[0-9]+(\.[0-9]+)?(-.*)?)$ ]]; then
+    elif [[ $1 =~ ^gdc-(([0-9]+\.[0-9]+(\.[0-9]+)?)([-+].*)?)$ ]]; then
         local name=${BASH_REMATCH[0]}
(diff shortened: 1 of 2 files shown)


Bobo1239:
Not really sure how we want to handle this. Versions higher than 5.2.0 (currently 5.4.0 and 6.3.0) are only available with the new url format but the new format additionally requires specifying the D language version (e.g. `4.8.2` -> `4.8.2+2.064.2`) which would break current users of the script. I don't know if there's some place where we could query the corresponding D version.
btw: ftp://ftp.gdcproject.org/LATEST is for some reason 4.8.5 instead of 6.3.0.

MartinNowak:
So far @ibuclaw has taken care of releasing stable gdc versions for those installers, so there is no new version/url format, just some releases were provided especially for the installer.
Last time I asked him he didn't intend to update that to the version on the website.
Also the gcc version didn't always increase in the past, last step was from 4.9.3+2.066.1 to 4.8.5+2.068.2.
Let's see whether 6.3.0 is something for broader adoption.

MartinNowak:
Will work on this with Iain soon on the GDC side.

lindt:
6.3.0 is released, but gdc still cannot be installed via install.sh

CyberShadow:
@ibuclaw Any news?

Bobo1239:
@CyberShadow I'm currently unsure what the future of GDC looks like as it was [upstreamed](https://www.phoronix.com/scan.php?page=news_item&px=GCC-9-Merges-D-Language) to GCC. According to the article it should become available in GCC 9.1 (end of Q1'2019).

CyberShadow:
OK, so distros will presumably package it then as part of the family of gcc packages.

What about Windows? Is the expectation on the Win32/POSIX toolkits (Cygwin/MSys2 etc.) to do the same?

ZombineDev:
Also, AFAIK, for the next 1-2 years the GDC in GCC svn will be based a dmd-cxx (~2.076 - static foreach) , while GDC master is based on a weekly dmd port, though @ibuclaw should correct me if I am wrong.

ibuclaw:
Strictly speaking, official releases are the only ones that matter. Anything that is earlier than gcc-9 should be ignored.

ZombineDev:
Official or not, the question is how someone can test a project against GDC based on e.g.  2.083, 2.082 and 2.080? For LDC this would be simply `install.sh ldc-1.13`, `install.sh ldc-1.12` and `install.sh ldc-1.10`.

ibuclaw:
There won't be a gdc release based on any of those versions. Will likely be 2.089 or 2.090 the next if dmd continues its rough time based release cycle.

I'll leave binaries up to distributions to make, as how gcc integrates varies per OS due to local patches, configure options, etc.

AndrewEdwards:
@ibuclaw where do we sit with this? GDC is has officially been in GCC for a while now. Can we move forward with this or can it completely abandoned? Not sure @Bobo1239 is even interested in what happens here anymore since. Hoping that he's still lurking around the community though.

What I would like to know is if there is any plan, now or ever, to see this to completion? If there isn't, there's no pint in keeping this request open. If there is, please state your timeline so that I can know when to reach out.

ibuclaw:
> @ibuclaw where do we sit with this? GDC is has officially been in GCC for a while now. Can we move forward with this or can it completely abandoned? 

Old binaries have been [archived off](https://gdcproject.org/downloads), though old links still work.

Personally, I prefer to leave binary releases to the package maintainers of various distributions.  If someone wants to take any of the CI scripts/infrastructure for gdc and use it to make binary packages out of, then be my guest and raise a pull request to add that functionality.

> how someone can test a project against GDC based on e.g. DMD 2.083, 2.082 and 2.080?

This question makes little sense as gcc and dmd release versions are mutually exclusive (there is no double-barrelled versioning).

AndrewEdwards:
CC: @PetarKirov @MartinNowak @CyberShadow

@Bobo1239, based on @ibuclaw's comments, there will be no packaging of GDC outside of the official GCC release. If we still intend to install GDC via the `install.sh` script, the DMD release management team will need to package and host the binaries themselves.

Should you choose to accept this responsibility, then this pull request remains valid. Otherwise, it will not be supported and, therefore, should be closed.

PR #252, opened 2017-09-07
install.sh: Fix -h handling and error on unrecognized parameters

CyberShadow:


script/install.sh
```diff
@@ -277,6 +277,11 @@ parse_args() {
             dmd | dmd-* | gdc | gdc-* | ldc | ldc-*)
                 compiler=$1
                 ;;
+
+            *)
+                usage
+                fatal "Unrecognized command-line parameter: $1"
```
wilzbach: Maybe first showing the error message first?

CyberShadow: No. I intentionally put it after the usage text. You're more likely to notice stuff at the bottom of your terminal window (near the prompt) than near the top.

wilzbach: Fair enough, I wouldn't have minded anyways

MartinNowak: Indeed, it's weird that most tools do it vice versa.

jondegenhardt:
This PR appears to have exposed a bug in the install script. The argument parsing for `-p | --path` (https://github.com/dlang/installer/blob/master/script/install.sh#L254-L259) take a second argument, the path. However, the '-p|--path' argument handler doesn't actually consume the argument, which then get run through arg processing loop. Previously it was ignored. Now the script exits.

I found this trying to run the docker-ldc script, uses the -p option: https://github.com/lindt/docker-ldc/blob/master/Dockerfile#L11

CyberShadow:
Looks like Martin had fixed this by the time you ran into it: 143bbf8669447f66f51402ac6db90de1350407ca

MartinNowak:
Was already wondering how that could have been there forever ;).

jondegenhardt:
And thanks for putting the error message at the bottom of the window!

PR #270, opened 2017-11-02
restrict installed configuration files to current OS

rainers:
It has always been bothering me that the windows installer also added configuration files for solaris, osx and others.


MartinNowak:
Feel free to merge such stuff, we'll eventually figure out when sth. breaks the builds.

PR #271, opened 2017-11-03
create dmd "light" installer on windows

rainers:
- grabs mingw from it's repository
- builds import libraries from mingw definition files
- builds stub for msvcr100.dll
- creates installer using these sdk libraries and lld-link

I couldn't test build_all though because it only runs on linux.

I'm also a bit confused about the use of the create_dmd_release_extra folder: it seems they do not end up in the result directory/zip. Instead link.exe is expected create_dmd_release\extras\dmd2\windows\bin, not in the additional extras. So what's the point of the committed extra folders?

This depends on https://github.com/dlang/tools/pull/267
The dmd/phobos sources should have https://github.com/dlang/dmd/pull/7261 and https://github.com/dlang/phobos/pull/5839 applied.

### DIFF
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -485,6 +485,9 @@ int main(string[] args)
         extract(cacheDir~"/"~libCurl, workDir~"/windows/old-dmd/");
         // Get updated OMF import libraries
         extract(cacheDir~"/"~omflibs, workDir~"/windows/old-dmd/dmd2/windows/lib/");
+
+        // grab mingw sources
+        run("git clone --depth 1 --branch 5.0-active https://git.code.sf.net/p/mingw/mingw-org-wsl.git " ~ workDir ~ "/clones/mingw");
     }
 
     cloneSources(gitTag, dubTag, isBranch, skipDocs, workDir~"/clones");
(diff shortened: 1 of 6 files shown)


MartinNowak:
Great, I'll try to test and review this for the upcoming beta.
The additional extras was necessary to copy over binary-only tools, like obj2asm.

MartinNowak:
Can we find a more descriptive name for this toolchain @rainers ? How about dmd-lld?
Also this should get a changelog entry, describing this (and an experimental warning I guess).

Just copy the changelog/README.md from dmd or so, installer changelogs are already supported.
https://github.com/dlang/tools/blob/40b1efb88550c6a2daea2b8c2d03177a502ce684/changed.d#L409

MartinNowak:
Should we try to integrate this with the normal dmd installer right away? Adding more choices might cause unnecessary confusion.

MartinNowak:
I'm for integrating this as a VC fallback with the normal release (for `-m32mscoff`, `-m64`).

rainers:
> I'm for integrating this as a VC fallback with the normal release (for -m32mscoff, -m64).

Done in #281. Closing in favor of that.

PR #272, opened 2017-11-25
Attempt to make windows installer more silent when /S is given

Manuzor:
This PR attempts to silence the installer when the /S option is passed. The installer opens messages boxes in certain circumstances. Most of them are just informative and do not actually require user intervention. For example, one of them occurs when no VS is installed and asks whether it should automatically install VS2013.

This obviously prevents fully automated and unattended installations of dmd using this installer. I've stumbled upon this because I'm currently setting up such a scenario and tried using that installer on a fresh Window install.

I have to say, however, that I was not able to build and test the installer myself because I didn't find documentation on how to set it up or even how to begin... So I'm kind of providing this change "blindly". I'd be glad to test the installer manually if I had some form of instructions on how to create it.

(I tried the bat file in the bootstrapper dir with the argument `v2.077.0` but it fails halfway through with a compile errors in `src\object.d`...)

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -144,7 +144,7 @@ SetCompressor /SOLID lzma
   System::Call 'kernel32::CreateMutexA(i 0, i 0, t "digital_mars_d_compiler_installer") ?e'
   Pop $R0
   StrCmp $R0 0 +3
-    MessageBox /SD IDOK MB_OK|MB_ICONSTOP "An instance of DMD installer is already running"
+    MessageBox MB_OK|MB_ICONSTOP "An instance of DMD installer is already running" /SD IDOK
     Abort
 !macroend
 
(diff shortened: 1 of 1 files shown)


atilaneves:
LGTM

MartinNowak:
Looks good @Manuzor, but please test silent install and uninstall, it's easy to overlook sth. with NSIS.

To build the installer, you need a windows zip build, e.g. just use http://downloads.dlang.org/releases/2.x/2.077.1/dmd.2.077.1.windows.7z, then run NSIS with the unpacked dmd2 folder like here (w/ `ver = 2.077.1`):
https://github.com/dlang/installer/blob/17936c860ea697d9fbd04f309990f51dced1babb/create_dmd_release/build_all.d#L328-L330.
This also needs a few NSIS plugins, listed here:
https://gist.github.com/MartinNowak/8270666#file-setup-ps1-L54-L67.

Manuzor:
Thanks @MartinNowak for the instructions. I've successfully created an installer and tested /S on two different Windows VMs (Windows 10 and Windows Server 2012). Seems to work very well!

However, I did not have to manually install any of the plugins you mentioned. Maybe they're automatically downloaded or are already included with NSIS by now?

From my point of view, this is ready to merge.

MartinNowak:
Thanks @Manuzor, this will land in 2.079.0, to be released on 2018-03-01 (beta starts on 2018-02-14).

PR #273, opened 2017-12-04
Use a fallback mirror for code.dlang.org

wilzbach:
As https://github.com/dlang/installer/pull/218 doesn't seem to be moving forward, I have uploaded all DUB releases to GitHub, which should be much more reliable (see https://github.com/dlang/dub/issues/1290 for details).
At the moment, there isn't a mirror for `code.dlang.org/download/LATEST` (we should probably create a gh-pages repo like LDC or use an S3 bucket), so for now I simply used the GitHub API. I know that it's rate-limited, but imho having a chance to succeed is a lot better than having none.

If code.dlang.org is down, we still manage to get the latest DUB:

```
dmd-2.077.1 already installed
curl: (7) Failed to connect to code.dlang.org port 80: Connection refused
curl: (7) Failed to connect to code.dlang.org port 80: Connection refused
curl: (7) Failed to connect to code.dlang.org port 80: Connection refused
curl: (7) Failed to connect to code.dlang.org port 80: Connection refused
curl: (7) Failed to connect to code.dlang.org port 80: Connection refused
Failed to download 'http://code.dlang.org/download/LATEST'
Downloading and unpacking http://code.dlang.org/files/dub-1.6.0-linux-x86_64.tar.gz

curl: (7) Failed to connect to code.dlang.org port 80: Connection refused
Downloading and unpacking https://github.com/dlang/dub/releases/download/v1.6.0/dub-1.6.0-linux-x86_64.tar.gz
######################################################################## 100.0%

Run `source ~/dlang/dmd-2.077.1/activate` in your shell to use dmd-2.077.1.
This will setup PATH, LIBRARY_PATH, LD_LIBRARY_PATH, DMD, DC, and PS1.
Run `deactivate` later on to restore your environment.
```

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -763,19 +763,25 @@ install_dub() {
         log "no dub binaries available for $OS"
         return
     fi
-    local url=http://code.dlang.org/download/LATEST
-    logV "Determining latest dub version ($url)."
-    dub="dub-$(fetch $url)"
+    local latestMirrors=(
+        "http://code.dlang.org/download/LATEST"
+        "http://dlang.github.io/dub/LATEST"
+    )
+    logV "Determining latest dub version (${latestMirrors[0]})."
+    dubVersion="$(fetch "${latestMirrors[@]}")"
+    dub="dub-${dubVersion}"
     if [ -d "$ROOT/$dub" ]; then
         log "$dub already installed"
         return
     fi
     local tmp url
     tmp=$(mkdtemp)
-    url="http://code.dlang.org/files/$dub-$OS-$ARCH.tar.gz"
-
-    log "Downloading and unpacking $url"
-    download_without_verify "$tmp/dub.tar.gz" "$url"
+    local mirrors=(
+        "http://code.dlang.org/files/$dub-$OS-$ARCH.tar.gz"
+        "https://github.com/dlang/dub/releases/download/${dubVersion}/$dub-$OS-$ARCH.tar.gz"
+    )
+    log "Downloading and unpacking ${mirrors[0]}"
+    download_without_verify "$tmp/dub.tar.gz" "${mirrors[@]}"
     tar -C "$tmp" -zxf "$tmp/dub.tar.gz"
     logV "Removing old dub versions"
     rm -rf "$ROOT/dub" "$ROOT/dub-*"


script/install.sh
```diff
@@ -659,24 +676,34 @@ list_compilers() {
     fi
 }
 
+find_latest_dub() {
+    local url version
+    url=http://code.dlang.org/download/LATEST
+    logV "Determining latest dub version ($url)."
+    version=$(fetch $url)
+    if [ -z "${version}" ] ; then
+        version=$(curl -s https://api.github.com/repos/dlang/dub/releases/latest | grep tag_name | sed -e 's/.*"tag_name": "v\([^"]*\)".*/\1/')
```
wilzbach: This will be replaced soon once an actual mirror for LATEST exists

wilzbach:
Are you okay with adding this fallback mirror @MartinNowak?

wilzbach:
What's wrong with this? Code.dlang.org is often down and we mirrored all dub releases to GitHub.
Also only DMD and LDC ship with dub

wilzbach:
A few things have changed here:
1) People are still using GDC and old DMD compilers and experience failures due to code.dlang.org being down
2) There's https://dlang.github.io/dub/LATEST now (see https://github.com/dlang/dub/issues/1290#issuecomment-357404749), so now workarounds will be needed
3) There's https://github.com/dlang/installer/pull/291 which adds mirroring capability to `fetch` and `download`

-> I'm reopening this, s.t. it doesn't get forgotten and I will revive and rebase it once https://github.com/dlang/installer/pull/291 has been merged. It will be a simple two-line diff then.

wilzbach:
> I will revive and rebase it once #291 has been merged.

Done. Also note that the DUB binaries [will soon be auto-deployed to GitHub](https://github.com/dlang/dub/issues/1367), s.t. code.dlang.org/download might be retired.

wilzbach:
Going to merge this now as the errors are still a thing and the backup "mirror" now got the primary "upstream" for the binaries.

script/install.sh
```diff
@@ -763,19 +763,25 @@ install_dub() {
         log "no dub binaries available for $OS"
         return
     fi
-    local url=http://code.dlang.org/download/LATEST
-    logV "Determining latest dub version ($url)."
-    dub="dub-$(fetch $url)"
+    local latestMirrors=(
+        "http://code.dlang.org/download/LATEST"
+        "http://dlang.github.io/dub/LATEST"
```
MartinNowak: Why not https?

wilzbach: I think simply because it wasn't using https before

MartinNowak:
Sorry, must have forgotten to submit a comment or accidentally hit the close button. Definitely helpful in the short-term.

It's still a bit unfortunate that dub releases are lingering between dlang.org and code.dlang.org (and now GH).
I'm not particularly convinced that we want to go further into GH releases as they are quite cumbersome and rely on quite a lot of services.

How about just archiving old dub releases that preceed the dmd integration.
I'm also open to extract separate binary dub tars/zips from the official releases and publish them on the existing release pipeline.

My main point here is to avoid the maintenance overhead of a second release pipeline when we already have established infrastructure and processes in place.

PR #274, opened 2017-12-04
Parse a configuration file

wilzbach:
I really like the installer, but I would prefer if it would use a different directory by default.
This PR allows the installer script to read a configuration file and thus have permanent configuration settings. I know that I could write a small wrapper around the installer or simply patch it, but I believe that I am not the only one having this problem.

tl;dr: With this PR, one can define `~/.config/dlang/installer.cfg` and do sth. like:

```
ROOT=~/.dlang
```

To use a different default directory.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -124,17 +124,19 @@ check_tools() {
 
 # ------------------------------------------------------------------------------
 
-mkdir -p "$ROOT"
-TMP_ROOT=$(mktemp -d "$ROOT/.installer_tmp_XXXXXX")
+init() {
+    mkdir -p "$ROOT"
+    TMP_ROOT=$(mktemp -d "$ROOT/.installer_tmp_XXXXXX")
 
-mkdtemp() {
-    mktemp -d "$TMP_ROOT/XXXXXX"
-}
+    mkdtemp() {
+        mktemp -d "$TMP_ROOT/XXXXXX"
+    }
 
-cleanup() {
-    rm -rf "$TMP_ROOT";
+    cleanup() {
+        rm -rf "$TMP_ROOT";
+    }
+    trap cleanup EXIT
 }
-trap cleanup EXIT
 
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


script/install.sh
```diff
@@ -124,17 +124,19 @@ check_tools() {
 
 # ------------------------------------------------------------------------------
 
-mkdir -p "$ROOT"
-TMP_ROOT=$(mktemp -d "$ROOT/.installer_tmp_XXXXXX")
+init() {
```
wilzbach: This needs to be done _after_ the configuration and CLI arguments are read.
Note that at the moment even if `--path` was set, the temporary directory was `~/dlang/.installer_tmp`.

MartinNowak:
Interesting, what other config options come to mind?
We should definitely fix the `--path` installation.

We could infer the install path from the install.sh location, seems ergonomically better than a config.

wilzbach:
> Interesting, what other config options come to mind?

`force-gpg` maybe?
`no-auto-update` (If you want to to do auto-updates of the installer, this would be an option too.

If we start to offer backup mirrors for the downloads, that could be a config as well.

> We could infer the install path from the install.sh location, seems ergonomically better than a config.

Well there are use cases where someone wants to use a different download folder for the binaries than the location of the script, but yes I think using the script's directory would be a very reasonable default.

MartinNowak:
A different path could easily be configured by symlinking ~/dlang to wherever you want.
I'd certainly say that `~/my/path/install.sh` should default to use `~/my/path`.
The rest of the config options seem vague, so a config file adds complexity and seems unwarranted atm.

wilzbach:
> A different path could easily be configured by symlinking ~/dlang to wherever you want.
> I'd certainly say that ~/my/path/install.sh should default to use ~/my/path.

We can't do that without breaking changes.
Currently the location is always:

```
ROOT=~/dlang
```

Who knows how people depend on this?

> The rest of the config options seem vague, so a config file adds complexity and seems unwarranted atm.

Always updating the installer (https://github.com/dlang/installer/pull/302) is something you requested yourself ;-)
And there we should give people an easy way to opt out. Of course, environment variables would do the job too, but they don't scale well.

PR #275, opened 2017-12-06
[WIP] Add AppVeyor build script

wilzbach:
Copy-pasting this script around, isn't very feasible.
The idea is to do sth. like this for now:

```yml
build_script:
- ps: wget http://dlang.org/install.ps -OutFile install.ps1
- ps: .\install.ps1
```

(we can always look into getting native AppVeyor support later)

This is far from ready, but I just configured AppVeyor for this repo, s.t. I can directly test this without needing a Windows machine.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 image:
 - Visual Studio 2015
-- Visual Studio 2017
+#- Visual Studio 2017
 platform: x64
 environment:
  matrix:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,62 @@
+image:
+- Visual Studio 2015
+- Visual Studio 2017
+platform: x64
+environment:
+ matrix:
+  - DC: dmd
+    DVersion: nightly
+    arch: x64
+  #- DC: dmd
+    #DVersion: nightly
+    #arch: x86
+  #- DC: dmd
+    #DVersion: beta
+    #arch: x64
+  #- DC: dmd
+    #DVersion: beta
+    #arch: x86
+  #- DC: dmd
+    #DVersion: stable
+    #arch: x64
+  #- DC: dmd
+    #DVersion: stable
+    #arch: x86
+  #- DC: dmd
+    #DVersion: 2.077.1
+    #arch: x64
+  #- DC: dmd
+    #DVersion: 2.077.1
+    #arch: x86
+  #- DC: ldc
+    #DVersion: beta
+    #arch: x86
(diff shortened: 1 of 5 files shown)


wilzbach:
A Windows CLI installer I just came across: https://flynn.io/docs/cli

MartinNowak:
That would be the way to go.

```ps
(New-Object Net.WebClient).DownloadString('https://dlang.org/install.psi') | iex
```

Not sure about options, maybe `iex -ArgumentList dmd-nightly`?

CyberShadow:
Generally you want to pin all dependencies' versions in CI. Using the latest version of something means any change could trigger it (if a new dependency release happened since the last change), and such breakages may happen at times when you don't have time to deal with the fallout.

MartinNowak:
You also want to test your project against the latest compiler without constantly updating configs for every small lib.

rmanthorpe:
I'd like to help out with this - I work on Windows every day and I'm familiar with PowerShell. It seems to me that this PR is conflating two separate issues:
1. Creating a PowerShell install script similar to install.sh
1. Adding an Appveyor setup script based on the install script.

One obviously depends on the other but a more general install script could be much more versatile - Travis for a start, any other CI too. I think the install script shouldn't make any assumptions about which visual studio version it should be using or anything like that - in the more general setting a user should be free to go on to configure visual studio however they like (or not at all).

wilzbach:
@rmanthorpe yes. The script comes from one of my Appveyor projects, so that's how it started.
However, as I haven't used Windows for a very long time, I basically lack the skill set (and motivation) to finish this, so any help would be more than welcome. I can help with reviewing and deploying this script.

As you said yourself, it's probably best to remove all the Visual Studio bits and focus on solely downloading the binaries in an "official" installer script.
Lastly, IIRC the download URL changed for LDC binaries at some point, so you might want to have a look at e.g. https://github.com/dlang-community/D-Scanner/blob/master/appveyor.yml

AndrewEdwards:
There's been no action on this since May 2019. @rmanthorpe, please carve out some of your time to lend a hand in bringing this to closure. If no do not have the time or inclination to work on this anymore, please state so. @wilzbach, please work with Richard to get this done. Alternatively, if the need for this no longer exists, please communicate that and close the request.

wilzbach:
Thanks a lot for the ping, but as the install script can now be called via cygwin it supports installation on AppVeyor and the other CI platforms out-of-the-box and doesn't require us to maintain two different scripts.

PR #276, opened 2017-12-12
gcstub was removed from druntime

MartinNowak:
- see dlang/druntime#1957


wilzbach:
Failure on AppVeyor is expected as the script is WIP -> https://github.com/dlang/installer/pull/275

PR #281, opened 2017-12-24
Windows installer: install mingw libs and improve VS selection

rainers:
- add mingw libraries to installation
- add installer page to select VS installation
- no need to patch sc.ini as dmd detects VS

@MartinNowak You will still have to add lld-link.exe to the extra folders.

The additional page only appears if no VS installation is detected and looks like this:

dinstaller

### DIFF
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -230,8 +230,32 @@ void prepareExtraBins(string workDir)
         "windbg.hlp", "ddemangle.exe", "lib.exe", "link.exe", "make.exe",
         "replace.exe", "shell.exe", "windbg.exe", "dm.dll", "eecxxx86.dll",
         "emx86.dll", "mspdb41.dll", "shcv.dll", "tlloc.dll", "libcurl.dll",
+        "lld-link.exe",
     ].addPrefix("bin/");
     auto winBins64 = ["libcurl.dll"].addPrefix("bin64/");
+    auto mingwCoffLibs = [
+        "aclui.lib", "advapi32.lib", "avicap32.lib", "avifil32.lib", "bthprops.lib",
+        "cap.lib", "comctl32.lib", "comdlg32.lib", "crypt32.lib", "ctl3d32.lib",
+        "d3d8.lib", "d3d9.lib", "d3dim.lib", "d3drm.lib", "d3dx8d.lib", "d3dx9d.lib",
+        "d3dxof.lib", "ddraw.lib", "dhcpcsvc.lib", "dinput8.lib", "dinput.lib", "dlcapi.lib",
+        "dnsapi.lib", "dplayx.lib", "dpnaddr.lib", "dpnet.lib", "dpnlobby.lib", "dpvoice.lib",
+        "dsetup.lib", "dsound.lib", "faultrep.lib", "gdi32.lib", "gdiplus.lib", "glaux.lib",
+        "glu32.lib", "icmui.lib", "igmpagnt.lib", "imagehlp.lib", "imm32.lib", "iphlpapi.lib",
+        "kernel32.lib", "ksproxy.lib", "ksuser.lib", "lz32.lib", "mapi32.lib", "mfcuia32.lib",
+        "mgmtapi.lib", "mprapi.lib", "mpr.lib", "mqrt.lib", "msacm32.lib", "mscms.lib",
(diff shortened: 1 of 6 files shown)


create_dmd_release/build_all.d
```diff
@@ -487,6 +487,9 @@ int main(string[] args)
         extract(cacheDir~"/"~libCurl, workDir~"/windows/old-dmd/");
         // Get updated OMF import libraries
         extract(cacheDir~"/"~omflibs, workDir~"/windows/old-dmd/dmd2/windows/lib/");
+
+        // grab mingw sources
+        run("git clone --depth 1 --branch 5.0-active https://git.code.sf.net/p/mingw/mingw-org-wsl.git " ~ workDir ~ "/clones/mingw");
```
wilzbach: How about checking out a specific tag or commit here to avoid breakage if the branch gets updated?

rainers: > How about checking out a specific tag or commit here to avoid breakage if the branch gets updated?

Maybe better, but 5.0-active is a release branch with only little activity, the real work seems to happen on 5.1-trunk.

wilzbach: Ah I just saw the `-active`

MartinNowak:
Looks great @rainers.

Saw it a little late, I'd rather target 2.079 with this change, since there isn't much time for testing the installer.

Maybe select and mark 2017 as `(default)`? A bit more information would be helpful to make that choice.

> You will still have to add lld-link.exe to the extra folders.

We do I get the binaries from? Should we handle this like curl?

rainers:
> Saw it a little late, I'd rather target 2.079 with this change, since there isn't much time for testing the installer.

Yeah, unfortunately this didn't get enough attention. So targetting next release is fine with me. A nightly build having the mingw extras would be nice. Maybe we can set an Appveyor task to build the installer.

> Maybe select and mark 2017 as (default)? 

It's already the default but moving it into first place, too, makes sense.

> We do I get the binaries from? Should we handle this like curl?

Where do you get these from? I see different sources:

* http://ftp.digitalmars.com
* http://downloads.dlang.org/other/
* or the previous released dmd

lld-link.exe can be made available for download similarily. IIRC you recently built it yourself. If you want, I can do that, too (I'm usually building from LLVM master).

MartinNowak:
> IIRC you recently built it yourself.

That was quite messy IIRC.

> If you want, I can do that, too (I'm usually building from LLVM master).

Even better, let's put the manual in the wiki and/or do it via a build script.

> Where do you get these from? I see different sources:

Then let's put it to http://downloads.dlang.org/other/
Do we have proper versions to build?

> Maybe we can set an Appveyor task to build the installer.

That doesn't really fit into an automated release building. A task to build lld would work fine though.

MartinNowak:
I'll look into automating the lld-link build in the next couple of days.

rainers:
> Even better, let's put the manual in the wiki and/or do it via a build script.

Building lld is described on the LLVM page: https://lld.llvm.org/windows_support.html. TBH it didn't work out right so simple for me, too, but I think that happened because I was trying to integrate it with my current working copy of LLVM.

> Then let's put it to http://downloads.dlang.org/other/

How about creating a new "extras" file there, the stuff in the semitwist file are pretty dated. I wouldn't mind if the 20 year old "debugger" in there would be removed, too.

> Do we have proper versions to build?

I'm currently using the latest in the hope a couple of issues are resolved. Last time I checked /MAP is not supported (/lldmap exists instead with different output format) and quotes in the library search path are not removed.

> A task to build lld would work fine though.

I suspect building LLD from scratch will take too long for an appveyor job.

rainers:
> I suspect building LLD from scratch will take too long for an appveyor job.

Just rebuilt locally with a quad-core mobile i7 and it took about 23 minutes without the cloning. Appveyor seems to have 2 cores, could be close for the timeout on appveyor (50 miutes IIRC).

rainers:
> I suspect building LLD from scratch will take too long for an appveyor job.

Takes about 40 minutes: https://github.com/dlang/installer/pull/286

create_dmd_release/build_all.d
MartinNowak: Let's use a cacheable tarball (for a fixed commit), just had lots of issues with sourceforge's git. Not too convinced that their service is reliable.

MartinNowak:
I see, this depends on #289 now.

rainers:
Thank for taking care of these.

PR #282, opened 2017-12-27
Add 'ldc-nightly' install target

JohanEngelen:
This PR installs ldc-nightly in e.g. "ldc-8c0abd52-20171222"
(DMD installs nightlies in e.g. "dmd-master-2017-12-27")

LDC does not have a LATEST_NIGHTLY file online, so I opted for just downloading the nightlies release page and grepping for what the binary package name should look like.

Tested on OSX.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -205,7 +205,7 @@ command_help() {
   dmd|gdc|ldc           latest version of a compiler
   dmd|gdc|ldc-<version> specific version of a compiler (e.g. dmd-2.071.1, ldc-1.1.0-beta2)
   dmd|ldc-beta          latest beta version of a compiler
-  dmd-nightly           latest dmd nightly
+  dmd|ldc-nightly       latest nightly version of a compiler
   dmd-2016-08-08        specific dmd nightly
 '
 
@@ -474,6 +474,17 @@ resolve_latest() {
             logV "Determining latest ldc-beta version ($url)."
             COMPILER="ldc-$(fetch $url)"
             ;;
+        ldc-nightly)
+            local ci_url=https://github.com/ldc-developers/ldc/releases/tag/CI
+            logV "Finding latest ldc-nightly binary package (at $ci_url)."
+            local package
+            package="$(fetch $ci_url | grep -o "ldc2-[0-9a-f]*-$OS-$ARCH-[0-9]*.tar.xz" | head -n 1)"
+            if [[ $package =~ ^ldc2-([0-9a-f]*)-$OS-$ARCH-([0-9]*)\.tar\.xz$ ]]; then
+                COMPILER="ldc-${BASH_REMATCH[1]}-${BASH_REMATCH[2]}"
+            else
+                fatal "Could not find ldc-nightly binaries (OS: $OS, arch: $ARCH)"
+            fi
+            ;;
         gdc)
             local url=http://gdcproject.org/downloads/LATEST
             logV "Determing latest gdc version ($url)."
@@ -538,6 +549,15 @@ install_compiler() {
         fi
 
         download_and_unpack_without_verify "$ROOT/$compiler" "$url"
+
+    # ldc-nightly: ldc-8c0abd52-20171222
+    elif [[ $1 =~ ^ldc-([0-9a-f]+)-([0-9]+) ]]; then
+        local package_hash=${BASH_REMATCH[1]}
+        local package_date=${BASH_REMATCH[2]}
+        local url="https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-$package_hash-$OS-$ARCH-$package_date.tar.xz"
+
+        # Install into 'ldc-8c0abd52-20171222' directory.
+        download_and_unpack "$url" "$ROOT/$1"
 
     # gdc-4.8.2, gdc-4.9.0-alpha1, gdc-5.2, or gdc-5.2-alpha1
     elif [[ $1 =~ ^gdc-([0-9]+\.[0-9]+(\.[0-9]+)?(-.*)?)$ ]]; then
--- a/script/install.sh
+++ b/script/install.sh
@@ -205,7 +205,8 @@ command_help() {
   dmd|gdc|ldc           latest version of a compiler
   dmd|gdc|ldc-<version> specific version of a compiler (e.g. dmd-2.071.1, ldc-1.1.0-beta2)
   dmd|ldc-beta          latest beta version of a compiler
-  dmd|ldc-nightly       latest nightly version of a compiler
+  dmd-nightly           latest dmd nightly
+  ldc-latest-ci         latest ldc CI build (with assertions enabled)
   dmd-2016-08-08        specific dmd nightly
 '
 
@@ -474,15 +475,15 @@ resolve_latest() {
             logV "Determining latest ldc-beta version ($url)."
             COMPILER="ldc-$(fetch $url)"
             ;;
-        ldc-nightly)
+        ldc-latest-ci)
             local url=http://thecybershadow.net/d/github-ldc
-            logV "Finding latest ldc-nightly binary package (at $url)."
+            logV "Finding latest ldc CI binary package (at $url)."
             local package
             package="$(fetch $url)"
             if [[ $package =~ ldc2-([0-9a-f]*)-$OS-$ARCH. ]]; then
                 COMPILER="ldc-${BASH_REMATCH[1]}"
             else
-                fatal "Could not find ldc-nightly binaries (OS: $OS, arch: $ARCH)"
+                fatal "Could not find ldc CI binaries (OS: $OS, arch: $ARCH)"
             fi
             ;;
         gdc)
@@ -550,7 +551,7 @@ install_compiler() {
 
         download_and_unpack_without_verify "$ROOT/$compiler" "$url"
 
-    # ldc-nightly: ldc-8c0abd52-20171222
+    # ldc-latest-ci: ldc-8c0abd52
     elif [[ $1 =~ ^ldc-([0-9a-f]+) ]]; then
         local package_hash=${BASH_REMATCH[1]}
         local url="https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-$package_hash-$OS-$ARCH.tar.xz"
--- a/script/install.sh
+++ b/script/install.sh
@@ -475,11 +475,11 @@ resolve_latest() {
             COMPILER="ldc-$(fetch $url)"
             ;;
         ldc-nightly)
-            local ci_url=https://github.com/ldc-developers/ldc/releases/tag/CI
-            logV "Finding latest ldc-nightly binary package (at $ci_url)."
+            local url=http://thecybershadow.net/d/github-ldc
+            logV "Finding latest ldc-nightly binary package (at $url)."
             local package
-            package="$(fetch $ci_url | grep -o "ldc2-[0-9a-f]*-$OS-$ARCH-[0-9]*.tar.xz" | head -n 1)"
-            if [[ $package =~ ^ldc2-([0-9a-f]*)-$OS-$ARCH-([0-9]*)\.tar\.xz$ ]]; then
+            package="$(fetch $url)"
+            if [[ $package =~ ^([0-9a-f]*)[[:space:]]([0-9]*)$ ]]; then
                 COMPILER="ldc-${BASH_REMATCH[1]}-${BASH_REMATCH[2]}"
             else
                 fatal "Could not find ldc-nightly binaries (OS: $OS, arch: $ARCH)"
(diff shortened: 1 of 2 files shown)


script/install.sh
```diff
@@ -421,6 +421,16 @@ resolve_latest() {
             logV "Determining latest ldc-beta version ($url)."
             COMPILER="ldc-$(fetch $url)"
             ;;
+        ldc-nightly)
+            local CIurl=https://github.com/ldc-developers/ldc/releases/tag/CI
+            logV "Finding latest ldc-nightly binary package (at $CIurl)."
+            local package="$(curl -s $CIurl | grep -o "ldc2-[0-9a-f]*-$OS-$ARCH-[0-9]*.tar.xz" | head -n 1)"
```
wilzbach: Shellcheck error: 

```
In script/install.sh line 427:
            local package="$(curl -s $CIurl | grep -o "ldc2-[0-9a-f]*-$OS-$ARCH-[0-9]*.tar.xz" | head -n 1)"
                  ^-- SC2155: Declare and assign separately to avoid masking return values.
```

wilzbach: See: https://github.com/koalaman/shellcheck/wiki/SC2155

wilzbach:
> Tested on OSX.

And now Linux ;-)
Too increase coverage, you might want to add `ldc-nightly` to the `travis.sh` test script.

script/install.sh
JohanEngelen: How do you want me to fix this?
`local -r` silences the warning; i don't particularly care about error codes, because we handle the error elsewhere. Still, it might be nice for the user to see the raw error.. ?

wilzbach: I think both `local -r` and using a separate line for the definition work though the latter seems to be the more prevalent style in this script, so I would opt for that one.

JohanEngelen: Done

wilzbach:
Thanks. As @MartinNowak is the author and "owner" of this install script, I would wait for his feedback/approval before merging. I can't deploy it anyways :/

script/install.sh
```diff
@@ -421,6 +421,17 @@ resolve_latest() {
             logV "Determining latest ldc-beta version ($url)."
             COMPILER="ldc-$(fetch $url)"
             ;;
+        ldc-nightly)
+            local CIurl=https://github.com/ldc-developers/ldc/releases/tag/CI
+            logV "Finding latest ldc-nightly binary package (at $CIurl)."
+            local package
+            package="$(curl -s $CIurl | grep -o "ldc2-[0-9a-f]*-$OS-$ARCH-[0-9]*.tar.xz" | head -n 1)"
```
ZombineDev: Hopefully we can replace this with a more robust way of getting the latest CI build, but we can live with it for now.

ZombineDev:
Nit: can you squash the "Fix [..] warning" commit into the first one?

script/install.sh
MartinNowak: The uploads are not sorted and old ones aren't deleted, so that results in a random artifact, not the latest one.
Who's updating the sha of the CI tag btw?

MartinNowak: Please use `fetch` instead of `curl` so there is some retry logic.

script/install.sh
```diff
@@ -421,6 +421,17 @@ resolve_latest() {
             logV "Determining latest ldc-beta version ($url)."
             COMPILER="ldc-$(fetch $url)"
             ;;
+        ldc-nightly)
+            local CIurl=https://github.com/ldc-developers/ldc/releases/tag/CI
```
MartinNowak: Weird variable name, how about `url` or `ci_url` instead?

script/install.sh
```diff
@@ -421,6 +421,17 @@ resolve_latest() {
             logV "Determining latest ldc-beta version ($url)."
             COMPILER="ldc-$(fetch $url)"
             ;;
+        ldc-nightly)
```
MartinNowak: I'd say we should at least also resolve `ldc-2017222` so it's possible to install a compiler from a specific date, quite useful for bisecting an issue.
I'm not sure whether you actually plan to archive your nightly builds.

JohanEngelen: I also don't know whether we'll archive the nightlies.
This is a new feature though, so please just merge this.

JohanEngelen:
Wait with merging, this does the wrong thing.

script/install.sh
JohanEngelen: (nice, you spotted the problem too)

script/install.sh
JohanEngelen: (they aren't "nightlies" btw, it's a new build after each git commit)

ZombineDev: Perhaps we should name it `ldc-latest-ci-build`?

ZombineDev:
@JohanEngelen as mentioned in https://github.com/ldc-developers/ldc/issues/1883#issuecomment-351197706, I think the best solution would be to make your CircleCI job push commits to https://github.com/ldc-developers/ldc-developers.github.io, updating a file like `LATEST_CI`.

MartinNowak:
Looking at this you might want to add `ldc-lts` before anything else ;).
Shouldn't the other one always be the latest commit on master?
How about `ldc-edge` resolving to `ldc-$(git describe)` (e.g. `ldc-1.7.0-beta1-18-gf86e33f1`).
Indeed it would be cleaner if you could push LATEST_EDGE/CI to that page or to the CI release page.

MartinNowak:
At best you keep the uploads for a week, so found issues can be reproduced.
Maybe directly using CircleCI's artifacts would be cleaner? Though they don't seem to provide any delete option.

wilzbach:
@JohanEngelen did you manage to write such a `LATEST_CI` file?

JohanEngelen:
@wilzbach Haven't looked further into this. It's on the backburner...

CyberShadow:
> Wait with merging, this does the wrong thing.

Check if it's possible to formulate a GitHub API query for this. Even if parsing the result with just bash is infeasible, having a dependency on jq seems reasonable.

wilzbach:
The GitHub API is heavily rate-limited. Using it will lead to failures on heavily used CIs like Travis.

CyberShadow:
Oh, because it will share limits with the IPs of CI services?

That makes sense, but scraping HTML is still fragile. CI services generally allow adding secrets to the account, so an API authentication token can be used there. (Or, just move the logic elsewhere...)

JohanEngelen:
I have completely stopped working on this btw. Feel free to take over.

CyberShadow:
I really don't know that much about LDC development.

I could write a tiny web service that returns the needed information from the GitHub API, if it helps. But, I can only guarantee that its availability will be as good as e.g. forum.dlang.org, which is probably around only 99%.

CyberShadow:
Cool. Will you do the other part if I do that?

wilzbach:
Just the return result that this PR queries from the GitHub API should be more than enough, e.g. `8c0abd52-20171222`
The latest stable and beta release versions get already written to GitHub pages by the LDC people, so this is purely about the latest nightly version.

CyberShadow:
It looks like there are two versions on https://github.com/ldc-developers/ldc/releases/tag/CI right now:

- ldc2-203abd4f-*-20181121.*
- ldc2-7a2a5987-*-20181125.*

Should the script return just the last such version string pair, then?

CyberShadow:
How's this? http://thecybershadow.net/d/github-ldc/

wilzbach:
Rebased & updated the URLs. Should be good to go now (:

script/install.sh
```diff
@@ -474,6 +474,17 @@ resolve_latest() {
...
+            local url=http://thecybershadow.net/d/github-ldc
+            logV "Finding latest ldc-nightly binary package (at $url)."
+            local package
+            package="$(fetch $url)"
+            if [[ $package =~ ^([0-9a-f]*)[[:space:]]([0-9]*)$ ]]; then
+                COMPILER="ldc-${BASH_REMATCH[1]}-${BASH_REMATCH[2]}"
```
JohanEngelen: This is going to always fail for Windows, and also when Linux and macOS end up finalizing the build on different dates. See current binaries at https://github.com/ldc-developers/ldc/releases/tag/CI

The different dates issue can be fixed by LDC's CI though.

wilzbach: Can't you just exclude the date from the filename then?
(The install script doesn't support Windows, so that's not a big concern for now).

JohanEngelen: @kinke Any particular reason for adding the date? If it's only to help humans a little bit, then let's drop it if we cannot get macos and linux to sync better.

kinke: Yeah it's just to aid humans. If I don't clean up the packages for a longer while, there may be hundreds of files, so the date suffix isn't that helpful then anyway.

script/install.sh
kinke: I don't like the term 'nightly' either. I guess for DMD they are indeed proper nightlies with no difference compared to regular builds. For us, they are the latest successful CI builds, which means a revision may not be available for all platforms (sporadic failures, timeouts...), not available at the same time (Windows x64/multilib and OSX packages taking quite a bit longer than the Linux x64 one), and particularly, they feature enabled LLVM+LDC assertions (well, currently not for Windows, due to AppVeyor slowness), increasing compile times.

script/install.sh
JohanEngelen: It's a bummer that we cannot automate the package removal process :(

JohanEngelen:
@wilzbach @CyberShadow  The date has been removed from the filename. See: https://github.com/ldc-developers/ldc/releases/tag/CI

CyberShadow:
>  The date has been removed from the filename.

Updated [script](http://thecybershadow.net/d/github-ldc/) to show only the hash.

> Edit: wait... how does Vlads script determine which is the newest....???

The release ID apparently doesn't change when new files are uploaded, so the data is always at /repos/ldc-developers/ldc/releases/8840125.

> Having a proxy do that (and cache it for some minutes or so) to bypass the rate limit would be nice, of course.

GitHub does not count 301 replies against the rate limit, which means the script will always provide the latest data with no rate limit.

kinke:
Can you show the source of the script? I only see the result and wonder how it works. It's almost certain that the Linux package is available earlier than the OSX one, so if it just returns the hash of the latest Linux package, the according OSX package may not be available yet. It needs to be per-platform (something like `?platform=linux-x86_64`).

JohanEngelen:
> Well we could theoretically have a LATEST_CI_<platform> file for all platforms.

@kinke That doesn't sound all that bad and would simplify things a lot I feel.

kinke:
Not for us, as 3 CI services need to be able to push to that repo, requiring a token to do so, which could even be abused by some malevolent PR etc.
I guess @CyberShadow's implementation just needs an additional handful of lines, taking a platform input param and filtering the artifact filenames before sorting by upload date and extracing the git hash.

CyberShadow:
I changed it to just dump all URLs, so that you can grep it as needed. Does that help?

JohanEngelen:
@CyberShadow Look at the CI release page now. The problem is getting the URL belonging to the _latest_ packages. (getting the urls is not the problem).

CyberShadow:
The GitHub server is replying with a "304 Not Modified" to a request with an "If-Modified-Since" of "Tue, 04 Dec 2018 22:50:54 GMT", for a resource that very obviously has been modified within the past week. Looks like a bug in their API.

CyberShadow:
Caching issue was fixed (switched to ETag-based caching at GitHub support's suggestion). What next, should I filter the assets to bin them by the hash and only show the last one per hash?

> Can you show the source of the script?

Yes, sorry, here it is:
http://thecybershadow.net/d/github-ldc/index.txt

Library used is https://github.com/milo/github-api + https://github.com/milo/github-api/pull/24.

JohanEngelen:
> only show the last one per hash?

I'd say only show the last one per arch.

CyberShadow:
Oops, right. How's this?

JohanEngelen:
Looks good to me :)

kinke:
Yeah, thx a lot, seems to work fine.

JohanEngelen:
OK, I've fixed the script. It now works on my mac and on a Ubuntu server too!

JohanEngelen:
rename `ldc-nightly` to `ldc-latest-ci`?  Or keep the name for uniformity with DMD's?

JohanEngelen:
@wilzbach @kinke @CyberShadow I'm happy with it, it works well for me. Good to merge?

kinke:
Yep, lgtm now.

wilzbach:
The script needs manual deployment as it needs to be signed, but I just did so.
It should be available on Travis etc. now, e.g. 

```yaml
d: ldc-latest-ci
```

wilzbach:
Also, I forgot to say thanks to everyone involved in this effort!

@JohanEngelen @kinke you might want to open a thread in the NG, s.t. people know about this and actually enable it in their CIs.

PR #284, opened 2017-12-29
appveyor: build windows installer

rainers:
no docs for now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,6 +24,8 @@ build_script:
         7z x bup.zip dm\bin\implib.exe > $null
         Start-FileDownload "http://semitwist.com/download/app/dmd-localextras.7z" -FileName extras.7z
         7z x extras.7z > $null
+        Start-FileDownload "http://nsis.sourceforge.net/mediawiki/images/c/c9/Inetc.zip" -FileName inetc.zip
+        7z x -o"c:\Program Files (x86)\NSIS" inetc.zip 
 
   - call "c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
   - set PATH=c:\projects\dmd2\windows\bin;c:\projects\dm\bin;%PATH%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,6 +38,8 @@ build_script:
 
   - cd installer\create_dmd_release
   - rdmd --build-only create_dmd_release.d
+  - set VCDIR=%VCINSTALLDIR%
+  - set SDKDIR=%WindowsSdkDir%
   - create_dmd_release.exe master --skip-docs --extras=..\..\localextras-windows --host-dmd=c:\projects\dmd2\windows\bin\dmd.exe
 
   - for /f %%f in (dmd.master.windows\dmd2\src\VERSION) do (set version=%%f)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,13 +17,13 @@ build_script:
   - cd c:/projects/
   - ps: |
         Start-FileDownload "http://downloads.dlang.org/releases/2.x/$Env:D_VERSION}/dmd.$Env:D_VERSION.windows.7z" -FileName dmd2.7z
-        7z x dmd2.7z > /dev/null
+        7z x dmd2.7z > $null
         Start-FileDownload "http://downloads.dlang.org/other/dmc857c.zip" -FileName dmc857.zip
-        7z x dmd2.zip > /dev/null
+        7z x dmd2.zip > $null
         Start-FileDownload "http://ftp.digitalmars.com/bup.zip" -FileName bup.zip
-        7z x bup.zip dm\bin\implib.exe > /dev/null
+        7z x bup.zip dm\bin\implib.exe > $null
         Start-FileDownload "http://semitwist.com/download/app/dmd-localextras.7z" -FileName extras.7z
-        7z x extras.7z > /dev/null
+        7z x extras.7z > $null
 
   - set PATH=c:\projects\dmd2\windows\bin;c:\projects\dm\bin;%PATH%
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,8 @@ build_script:
   - ps: |
         Start-FileDownload "http://downloads.dlang.org/releases/2.x/$Env:D_VERSION/dmd.$Env:D_VERSION.windows.7z" -FileName dmd2.7z
         7z x dmd2.7z > $null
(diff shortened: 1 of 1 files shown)


wilzbach:
> Could you reenable appveyor for this repo? Thanks.

Done :)

appveyor.yml
```diff
@@ -0,0 +1,55 @@
...
+
+cache:
+
+init:
+  - git config --global core.autocrlf input
+
+build_script:
+  - cd c:/projects/
+  - ps: |
+    Start-FileDownload "http://downloads.dlang.org/releases/2.x/$Env:D_VERSION}/dmd.$Env:D_VERSION.windows.7z" -FileName dmd2.7z
```
wilzbach: BTW instead of copy/pasting this all over, we are [working towards a windows variant of the installer script](https://github.com/dlang/installer/pull/275/files). It got a bit stalled though as I'm no expert on Windows nor user.

rainers: > BTW instead of copy/pasting this all over, we are working towards a windows variant of the installer script. It got a bit stalled though as I'm no expert on Windows nor user.

I also noticed after creating this PR. Downloading the compiler is usually a smaller problem, but tedious.

rainers:
> Done :)

Thanks.

appveyor.yml
```diff
@@ -0,0 +1,60 @@
...
+  - ps: |
+        Start-FileDownload "http://downloads.dlang.org/releases/2.x/$Env:D_VERSION/dmd.$Env:D_VERSION.windows.7z" -FileName dmd2.7z
+        7z x dmd2.7z > $null
+        Start-FileDownload "http://downloads.dlang.org/other/dm857c.zip" -FileName dm857c.zip
+        7z x dm857c.zip > $null
+        Start-FileDownload "http://ftp.digitalmars.com/bup.zip" -FileName bup.zip
+        7z x bup.zip dm\bin\implib.exe > $null
+        Start-FileDownload "http://semitwist.com/download/app/dmd-localextras.7z" -FileName extras.7z
+        7z x extras.7z > $null
+        Start-FileDownload "http://nsis.sourceforge.net/mediawiki/images/c/c9/Inetc.zip" -FileName inetc.zip
```
wilzbach: One experience I have from maintaining D's CIs is that the network isn't 100% reliable, especially the FreeBSD Server from digitalmars.
If we can't put those things on downloads.dlang.org, it be be worthwhile to push them into a GitHub repository as mirror. This would have the additional advantage of being a mirror in our control.

appveyor.yml
```diff
@@ -0,0 +1,60 @@
...
+  - set PATH=c:\projects\dmd2\windows\bin;c:\projects\dm\bin;%PATH%
+
+  - if not exist build\nul md nul
+  - set BASE=%TEMP%\create_dmd_release
+  - git clone --depth 1 --branch %BRANCH% https://github.com/dlang/dmd.git "%BASE%/dmd"
+  - git clone --depth 1 --branch %BRANCH% https://github.com/dlang/druntime.git "%BASE%/druntime"
+  - git clone --depth 1 --branch %BRANCH% https://github.com/dlang/phobos.git "%BASE%/phobos"
+  - git clone --depth 1 --branch %BRANCH% https://github.com/dlang/tools.git "%BASE%/tools"
+  - git clone --depth 1 --branch %BRANCH% https://github.com/dlang/dlang.org.git "%BASE%/dlang.org"
+  - git clone --depth 1 --branch %BRANCH% https://github.com/dlang/dub.git "%BASE%/dub"
```
wilzbach: Could have been a loop and/or function calls to repeat in case of network failures, e.g. https://github.com/dlang/dmd/blob/master/.circleci/run.sh

Of course, this isn't needed for this initial version.

rainers:
> Do you plan to do more in this PR or should we squash and merge?

Not really, but I'm not sure this should go live for now:

- the localextras are pretty dated, they should be updated or replaced by a new package on downloads.dlang.org
- @MartinNowak could be right that this isn't really that helpful for generating releases, e.g it won't trigger nightlies for dmd updates.
- Martin might not be able to use these builds to do an actual release because it doesn't easily integrate with his workflow.

appveyor.yml
rainers: I've kept that task for your installer helper script https://github.com/dlang/installer/pull/275 ;-)

wilzbach: We don't need to clone the repos for the install script

wilzbach:
I think it's still a good step forward to merge this, because it will provide CI feedback for people improving the installer in the future.
Currently we don't have that "many" tests.

appveyor.yml
```diff
@@ -0,0 +1,60 @@
...
+build_script:
+  - cd c:/projects/
+  - ps: |
+        Start-FileDownload "http://downloads.dlang.org/releases/2.x/$Env:D_VERSION/dmd.$Env:D_VERSION.windows.7z" -FileName dmd2.7z
+        7z x dmd2.7z > $null
+        Start-FileDownload "http://downloads.dlang.org/other/dm857c.zip" -FileName dm857c.zip
+        7z x dm857c.zip > $null
+        Start-FileDownload "http://ftp.digitalmars.com/bup.zip" -FileName bup.zip
+        7z x bup.zip dm\bin\implib.exe > $null
+        Start-FileDownload "http://semitwist.com/download/app/dmd-localextras.7z" -FileName extras.7z
```
MartinNowak: That's hasn't been used in ages, you can take the binaries from the old release.

MartinNowak:
If we're going in the direction of testing the build script, we should figure out how to make the VBox build in build_all optional.
Since the build script is rarely update and almost exclusively breaks due to the other repos, I'm not sure how useful testing it would be.

If on the other hand we just want to test the installer, we should download an already released dmd.windows.7z and build the installer from that. This could even be the nightly https://nightlies.dlang.org/dmd-nightly/dmd.master.windows.7z in case new files need to be shipped.

appveyor.yml
rainers: > That's hasn't been used in ages, you can take the binaries from the old release.

How do you get anything new into that, e.g. link.exe or snn.lib?

wilzbach:
> Since the build script is rarely update and almost exclusively breaks due to the other repos, I'm not sure how useful testing it would be.

I think it's very useful, because this gives reviewers and contributors a _informative feedback_ when doing changes.

> Since the build script is rarely update and almost exclusively breaks due to the other repos, I'm not sure how useful testing it would be.

Fair point, but a bit unrelated, but maybe @braddr can add a new section "test install" to DAutoTest?

rainers:
Now on Azure.

PR #286, opened 2017-12-30
[experimental] build lld on appveyor

rainers:


rainers:
closing for now, trying it on a different account

rainers:
Seems to work now.

windows/build_lld.bat
```diff
@@ -0,0 +1,23 @@
+@setlocal
+
+set LLVM_BRANCH=master
+
+call "c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars32.bat"
+git clone --depth 1 --branch %LLVM_BRANCH% https://github.com/llvm-mirror/llvm.git llvm || exit /B 1
+git clone --depth 1 --branch %LLVM_BRANCH% https://github.com/llvm-mirror/lld.git llvm\tools\lld || exit /B 1
```
MartinNowak: I'd rather go with the released sources, then we also have proper versioning.
http://releases.llvm.org/download.html#5.0.1

http://releases.llvm.org/5.0.1/lld-5.0.1.src.tar.xz
http://releases.llvm.org/5.0.1/llvm-5.0.1.src.tar.xz

Bonus, check gpg signatures locally and put the sha256sums in the script (prolly simpler than gpg on window).

```
d5b36c0005824f07ab093616bdff247f3da817cae2c51371e1d1473af717d895  lld-5.0.1.src.tar.xz
5fa7489fc0225b11821cab0362f5813a05f2bcf2533e8a4ea9c9c860168807b0  llvm-5.0.1.src.tar.xz
```

windows/build_lld.bat
```diff
@@ -0,0 +1,23 @@
...
+set CMAKE_OPT=-G "Visual Studio 15"
+set CMAKE_OPT=%CMAKE_OPT% -DCMAKE_BUILD_TYPE=Release
+set CMAKE_OPT=%CMAKE_OPT% -DLLVM_TARGETS_TO_BUILD=X86 
+set CMAKE_OPT=%CMAKE_OPT% -DLLVM_INCLUDE_DIRS="c:/projects/llvm/include"
+
+cmake %CMAKE_OPT% ..\llvm || exit /B 1
+devenv LLVM.sln /project lld /Build "MinSizeRel|Win32" || exit /B 1
+
+if not exist ..\bin\nul md ..\bin
+copy MinSizeRel\bin\lld-link.exe ..\bin
```
MartinNowak: Compress the exe and output with a version number, e.g. `lld-link-5.0.1.zip`.

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+environment:
+  matrix:
+    - ARCH:                x86
+      BRANCH:              master
+
+cache:
+  - bin\lld-link.exe -> windows\build_lld.bat
+
+artifacts:
+  - path: bin\lld-link.exe
```
MartinNowak: Not sure whether you could use variables here `bin\lld-link-%VERSION%.zip`.

windows/build_lld.bat
```diff
@@ -0,0 +1,23 @@
...
+if not exist %lld_build_dir%\nul md %lld_build_dir%
+cd %lld_build_dir%
+
+set CMAKE_OPT=-G "Visual Studio 15"
+set CMAKE_OPT=%CMAKE_OPT% -DCMAKE_BUILD_TYPE=Release
+set CMAKE_OPT=%CMAKE_OPT% -DLLVM_TARGETS_TO_BUILD=X86 
+set CMAKE_OPT=%CMAKE_OPT% -DLLVM_INCLUDE_DIRS="c:/projects/llvm/include"
+
+cmake %CMAKE_OPT% ..\llvm || exit /B 1
+devenv LLVM.sln /project lld /Build "MinSizeRel|Win32" || exit /B 1
```
MartinNowak: Can we trigger this manually, or build the script when the version/file changed? Running half an hour for every unrelated change seems wasteful.

wilzbach: AppVeyor supports crons, so we could run it once a day?

appveyor.yml
rainers: According to the docs `lld-link-$(VERSION).zip` should work.

windows/build_lld.bat
rainers: The build can be triggered manually if you login to Appveyor. LLD is cached (from master builds) and is only rebuilt if the dependency changes, i.e. currently the batch file.

windows/build_lld.bat
rainers: Building a zip file now. I'm not sure compressed exes are so great, I suspect it makes program execution slower than when reading the program from a fast disk or the file cache.

windows/build_lld.bat
rainers: Changed to download the tarballs. How would you verify the SHA? (I don't think this is super important, though.)

windows/build_lld.bat
MartinNowak: For a fully manual build, you could add.
```yaml
branches: { only: [manually-triggered-build] }
```
But indeed using the cache is also a good idea.

windows/build_lld.bat
```diff
@@ -0,0 +1,30 @@
...
+
+zip ../../lld-link-%LLVM_VER%.zip MinSizeRel\bin\lld-link.exe
```
MartinNowak:
```sh
cd MinSizeRel\bin
7z a ..\..\..\..\lld-link-%LLMV_VER%.zip lld-link.exe
```

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+environment:
+  matrix:
+    - ARCH:                x86
+      LLVM_VER:            5.0.1
+
+cache:
+  - 'lld-link-%LLVM_VER%.zip -> windows\build_lld.bat'
+
+artifacts:
+  - path: bin\lld-link-$(LLVM_VER).zip
```
MartinNowak: without the `bin\` now

MartinNowak: And I guess it's also using `%LLVM_VER%` expansion.

MartinNowak:
AppVeyor would still copy the artifacts for every commit on master.
Let's simply make a couple of inactive `build-X` branches for the different builds, then we can manually trigger them and we have a clean history of artifacts.

rainers:
> AppVeyor would still copy the artifacts for every commit on master.

Artifacts could also be uploaded when rebuilt, see https://github.com/dlang/installer/pull/284/files#diff-180360612c6b8c4ed830919bbb4dd459R58 for an example.
But it might be a bit troublesome to find the last build in this case, so another branch sounds good to me, too.

MartinNowak:
Signed and uploaded
http://ftp.digitalmars.com/

rainers:
> Signed and uploaded

Thanks. Should we add a similar build branch for the mingw libraries? The build tool https://github.com/dlang/tools/pull/267 could be moved to the installer in that case.

MartinNowak:
Yes, let's add all the irregularly build ones.

rainers:
> Yes, let's add all the irregularly build ones.

https://github.com/dlang/installer/pull/289

PR #287, opened 2017-12-30
add small deploy script for install.sh

MartinNowak:


script/deploy.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -uexo pipefail
+
+${GPG:-gpg2} --detach-sign install.sh
```
wilzbach: So this still means you are the only one who can sign this? We should probably think about expanding our keyring...

script/deploy.sh
```diff
@@ -0,0 +1,9 @@
...
+scp install.sh{,.sig} digitalmars.com:/usr/local/www/dlang.org/data/
+scp install.sh{,.sig} nightlies.dlang.org:/var/www/builds/
+aws --profile ddo s3 cp install.sh s3://downloads.dlang.org/other/
+aws --profile ddo s3 cp install.sh.sig s3://downloads.dlang.org/other/
```
wilzbach: Lots of permissions needed here.

script/deploy.sh
jacob-carlborg: The D language should be responsible for this, all passwords and similar. So we don't depend on a single person.

PR #289, opened 2018-01-01
Build mingw COFF libraries

rainers:
includes https://github.com/dlang/tools/pull/267


MartinNowak:
http://downloads.dlang.org/other/mingw-libs-5.0.2.zip

PR #291, opened 2018-01-07
Use fallback mirrors in the install script

wilzbach:
Failures are getting more frequent, see e.g. https://github.com/travis-ci/travis-build/pull/1282
This is a first step at adding more mirrors.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -62,26 +62,43 @@ retry() {
         fi
     done
 }
-
 # url, path, [gpg signature url to verify]
+# url and the gpg signature can be an array of urls
 download() {
-    local url path
-    url=$1
-    path=$2
+    local mirrors path
+    mirrors=($1)
+    path="$2"
 
-    retry curl "$url" -o "$path"
+    try_all_mirrors() {
+        for i in "${!mirrors[@]}" ; do
+            if [ "$i" -gt 0 ] ; then
+                log "Falling back to mirror: ${mirrors[$i]}"
+            fi
+            if curl "${mirrors[$i]}" -o "$path" ; then
+                break
+            fi
+        done
+    }
+    retry try_all_mirrors
     if [ ! -z "${3:-}" ]; then
         verify "$3" "$path"
     fi
 }
 
-# url
+# url (can be an array of urls)
 fetch() {
-    local url path
-    url=$1
+    local mirrors path
+    mirrors=($1)
     path=$(mktemp "$TMP_ROOT/XXXXXX")
 
-    retry curl2 -sS "$url" -o "$path"
+    try_all_mirrors() {
+        for mirror in "${mirrors[@]}" ; do
+            if curl2 -sS "$mirror" -o "$path" ; then
+                break
+            fi
+        done
+    }
+    retry try_all_mirrors
     cat "$path"
     rm "$path"
 }
@@ -371,13 +388,17 @@ Run \`deactivate\` later on to restore your environment."
 }
 
 install_dlang_installer() {
-    local url tmp
-    url=https://dlang.org/install.sh
+    local tmp mirrors
     tmp=$(mkdtemp)
+    mirrors=(
+        "https://dlang.org/install.sh"
+        "https://downloads.dlang.org/other/install.sh"
+        "https://nightlies.dlang.org/install.sh"
+    )
 
     mkdir -p "$ROOT"
-    log "Downloading $url"
-    download "$url" "$tmp/install.sh" "$url.sig"
+    log "Downloading ${mirrors[0]}"
+    download "${mirrors[*]}" "$tmp/install.sh" "${mirrors[*]/%/.sig}"
     mv "$tmp/install.sh" "$ROOT/install.sh"
     rmdir "$tmp"
     chmod +x "$ROOT/install.sh"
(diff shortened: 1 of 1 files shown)


script/install.sh
```diff
@@ -62,26 +62,43 @@ retry() {
...
-    local url path
-    url=$1
+    local mirrors path
+    mirrors=($1)
     path=$(mktemp "$TMP_ROOT/XXXXXX")
 
-    retry curl2 -sS "$url" -o "$path"
+    try_all_mirrors() {
+        for mirror in "${mirrors[@]}" ; do
+            if curl2 -sS "$mirror" -o "$path" ; then
```
wilzbach: This is copy due to `curl2` being used here and `curl` in download. Is there any reason not to use `curl2` in  `download`?

script/install.sh
```diff
@@ -371,13 +388,17 @@ Run \`deactivate\` later on to restore your environment."
...
 
 install_dlang_installer() {
-    local url tmp
-    url=https://dlang.org/install.sh
+    local tmp mirrors
     tmp=$(mkdtemp)
+    mirrors=(
+        "https://dlang.org/install.sh"
+        "https://downloads.dlang.org/other/install.sh"
+        "https://nightlies.dlang.org/install.sh"
```
wilzbach: I would add `https://github.com/dlang/installer/raw/stable/script/install.sh` to the list too, but I wasn't sure whether you would be okay with it.

script/install.sh
```diff
@@ -540,13 +574,17 @@ download_and_unpack() {
 }
 
 verify() {
+    local keyring_mirrors=(
+        "https://dlang.org/d-keyring.gpg"
+        "https://github.com/dlang/dlang.org/raw/846269e913114fd77af0cc1aff67269046b70db8/d-keyring.gpg"
```
wilzbach: I couldn't find another mirror for this.

wilzbach:
FWIW for gdc downloads (e.g. at dlang/dmd) I still see failures related to the dub-registry being down often, e.g. https://semaphoreci.com/cybershadow/dmd/branches/pull-request-7659/builds/2, as gdc doesn't ship with dub.

So it would be cool to add a mirror for the DUB binaries too (see https://github.com/dlang/installer/pull/273)

script/install.sh
```diff
@@ -390,14 +411,20 @@ Run \`$ROOT/install.sh --help\` for usage information.
...
         dmd-beta)
-            local url=http://downloads.dlang.org/pre-releases/LATEST
-            logV "Determing latest dmd-beta version ($url)."
-            COMPILER="dmd-$(fetch $url)"
+            local mirrors=(
+                "http://downloads.dlang.org/pre-releases/LATEST"
+                "http://ftp.digitalmars.com/LATEST_BETA"
+            )
+            logV "Determing latest dmd-beta version (${mirrors[0]})."
+            COMPILER="dmd-$(fetch "${mirrors[*]}")"
```
MartinNowak: Think this should be `[@]` expanded, so fetch really gets 2 arguments.
```bash
cnt() { echo $#; }
args=('entry with' 'spaces')
cnt "${args[*]}" # 1
cnt "${args[@]}" # 2
cnt ${args[*]} # 3
cnt ${args[@]} # 3
```
As with "$@" you always want to pass arguments in an array using `"${ary[@]}".

script/install.sh
```diff
@@ -62,26 +62,43 @@ retry() {
...
     fi
 }
 
-# url
+# url (can be an array of urls)
 fetch() {
-    local url path
-    url=$1
+    local mirrors path
+    mirrors=($1)
```
MartinNowak: This does IFS/whitespace splitting. Rather pass all the arguments separately on the call-side and use `("$@")` here.
```sh
cnt() { echo $#; }
bug() { ary=($@); cnt "${ary[@]}"; }
bug 1 2 # -> 2
bug '1 1' 2 # -> 3 :/
ok() { ary=("$@"); cnt "${ary[@]}"; }
ok 1 2 # -> 2
ok '1 1' 2 # -> 2 :)
```

script/install.sh
```diff
@@ -62,26 +62,43 @@ retry() {
         fi
     done
 }
-
 # url, path, [gpg signature url to verify]
+# url and the gpg signature can be an array of urls
```
MartinNowak: Let's change it to `path, verify, url...` then, and always require 0 or 1 for verify.

MartinNowak: The signature can be assumed to be path+`.sig`.

script/install.sh
MartinNowak: Let's not automatically deploy this, it's not worth the security implications.
Even without sudo (which often is cached), `rm -rf ~` can do a lot of damage.

script/install.sh
```diff
@@ -371,13 +388,17 @@ Run \`deactivate\` later on to restore your environment."
...
+        "https://dlang.org/install.sh"
+        "https://downloads.dlang.org/other/install.sh"
+        "https://nightlies.dlang.org/install.sh"
+    )
 
     mkdir -p "$ROOT"
-    log "Downloading $url"
-    download "$url" "$tmp/install.sh" "$url.sig"
+    log "Downloading ${mirrors[0]}"
+    download "${mirrors[*]}" "$tmp/install.sh" "${mirrors[*]/%/.sig}"
```
MartinNowak: Again `.sig` is implicit when `verify=1`, so the call would be.
```sh
download "$tmp/install.sh" 1 "${mirrors[@]}"
```
Or for better readability, use `download_verify` and plain `download`.

script/install.sh
wilzbach: I'm aware (I did it on purpose after all) - urls aren't supposed to contain spaces :wink: 
I did have this, s.t. I can pack the urls for mirrors and their signatures as individual arguments for the `download` function.

script/install.sh
wilzbach: Ok

script/install.sh
MartinNowak: There you go, https://nightlies.dlang.org/d-keyring.gpg.
In fact you can upload files to sebastian@nightlies.dlang.org:/var/www/builds yourself.
But you cannot move or delete my entries in that folder (sticky bit). Allowing any nginx group member to mess with the install script and the signature key sounds like disaster.

MartinNowak:
LMostlyGTM

wilzbach:
Ping @MartinNowak - are we good here?
People are still seeing random failures on their CI jobs (e.g. https://travis-ci.org/d-widget-toolkit/dwt/builds/334451034). Note in that case the DUB binary failed, but I will revive https://github.com/dlang/installer/pull/273 as soon as the mirroring functionality has been merged. with this PR #273 will be a simple two-line diff.

CC @jacob-carlborg

script/install.sh
```diff
@@ -62,26 +62,54 @@ retry() {
...
-# url
+# path, urls...
+download_with_verify() {
+    download "$1" 1 "${@:2}"
+}
+
+# path, urls...
+download_without_verify() {
+    download "$1" 0 "${@:2}"
+}
```
wilzbach: I added these to functions to avoid the need of passing around a magic `1` or `0`

script/install.sh
```diff
@@ -536,20 +587,39 @@ download_and_unpack() {
...
+
+# path, urls...
+download_and_unpack_with_verify() {
+    download_and_unpack "$1" 1 "${@:2}"
+}
+
+# path, urls...
+download_and_unpack_without_verify() {
+    download_and_unpack "$1" 0 "${@:2}"
```
wilzbach: These two functions were added to avoid passing around magic `0` or `1`

wilzbach:
Ping @MartinNowak - just saw another random failure on Travis :/

PR #295, opened 2018-01-25
Copy the generated man pages to the release archive

wilzbach:
See also: https://github.com/dlang/dmd/pull/7669

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -514,6 +514,8 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
                 run(makecmd~" DOC_OUTPUT_DIR="~origDir~"/docs release");
                 // put into docs/ folder which gets copied to all other platforms
                 copyFile("d-tags-release.json", origDir~"/docs/d-tags-release.json");
+                // copy generated man pages to docs/man which gets copied to all other platforms
+                copyDir(cloneDir~"/dmd/generated/docs/man", origDir~"/docs/man");
             }
         }
         else version (Windows)
@@ -591,6 +593,7 @@ void createRelease(string branch)
             !a.startsWith("images/original/") &&
             !a.startsWith("chm/") &&
             ( a.endsWith(".html") || a.startsWith("css/", "images/", "js/") );
+        // copy docs from linux build
         copyDir(origDir~"/docs", releaseDir~"/dmd2/html/d", a => dlangFilter(a));
         version(Windows)
         {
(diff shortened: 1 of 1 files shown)


MartinNowak:
Still getting some nonsense output from dlang.org.
```
mkdir -p /home/vagrant/docs/docs/man/man1/
cp ../dmd/generated/docs/man/man1/dmd.1 /home/vagrant/docs/docs/man/man1/dmd.1
# CircleCi + nightlies.dlang.org might not have `man` installed
if command -v man > /dev/null ; then \
		make -s -C ../dmd/docs DMD=/home/vagrant/clones/dlang.org/.generated/stable_dmd-2.077.1/dmd2/linux/bin32/dmd DIFFABLE= preview > /home/vagrant/docs/docs/man/man1/dmd.txt; \
	fi
<standard input>:139: warning: numeric expression expected (got `s')
```

Why is the man page copied to the web folder, for diffing?
Dlang.org is also the only reason why the man page is build, seems a bit hacky.

Furthermore the dependency `$(MAN_PAGE)` on the preview target does not exist
https://github.com/dlang/dmd/blob/18ba25dcf81edfc437c54ee6b5445489d395500a/docs/Makefile#L25

and path and variable name are inconsistent between https://github.com/dlang/dlang.org/blob/4e07e9f5da92c45329bdcfb8d0195b4084ec1042/posix.mak#L349 (MAN_PAGE=docs/man/man1/dmd.1) and https://github.com/dlang/dmd/blob/18ba25dcf81edfc437c54ee6b5445489d395500a/docs/Makefile#L5 (DMD_MAN_PAGE=man/man1/dmd.1).

MartinNowak:
The man page is a little OS specific (-fPIC et.al.), but that should be fine for now.
Maybe make the generated man page generic for all OSes if you feel that's worth the effort.

wilzbach:
Thanks!

> Still getting some nonsense output from dlang.org.

Interesting, but according to https://lists.gnu.org/archive/html/groff/2008-02/msg00071.html it's harmless :O 

> Why is the man page copied to the web folder, for diffing?

Yes, s.t. changes can be previewed easily.

> Dlang.org is also the only reason why the man page is build, seems a bit hacky.

Oh that wasn't on purpose. It's built on the `install` target too - I thought that one is run here.
Should I add a separate man target to the `posix.mak` or should we call the makefile in `docs` directly?

> Maybe make the generated man page generic for all OSes if you feel that's worth the effort.

I don't think that's worth it.

> Furthermore the dependency $(MAN_PAGE) on the preview target does not exist

Oh sorry -> https://github.com/dlang/dmd/pull/7776

> and path and variable name are inconsistent between 

Yeah that was a bit on purpose.
For dlang.org it was nice to do the following

```
$W/$(MAN_PAGE): $(DMD_DIR)/generated/$(MAN_PAGE)
```

Well, if my [makefile unification](https://github.com/dlang/dmd/pull/7476),  ever gets approved, I'm happy to incorporate the man page variable into it.

PR #297, opened 2018-02-15
Add '--dmd' and '--dc' to print the path of the installed compiler

wilzbach:
This has come up frequently in the past.

Use case: one has a generic compiler variable like `$DMD` in a CI script and wants to use the install script without activating the environment.

See also:
- https://github.com/dlang/dmd/pull/7848#issuecomment-364235083
- https://github.com/dlang/tools/pull/297/files#diff-f9a62833538ab14fff849c2be16ab986R9

I'm not fully convinced about the name, so bike-shedding is allowed.

An alternative to this would have been to add a flag similar to `-a|--activate`.
That would require a two line change and would be fine for me too.

CC @WebDrake

### DIFF
--- a/changelog/install-dmd-dc.dd
+++ b/changelog/install-dmd-dc.dd
@@ -1,0 +1,39 @@
+Installation script now supports `get-path <compiler>`
+
+`get-path <compiler>` has been added as a new action to the install script.
+This action allows printing information about a compiler binary.
+The behavior of `get-path` can be changed with the following additional options:
+
+$(UL
+    $(LI `--install` install compiler if not found locally)
+    $(LI `--dmd` returns the path of a specific compiler executable (DMD-alike interface))
+    $(LI `--dub` returns the path of the DUB executable)
+)
+
+$(H4 Examples)
+
+$(CONSOLE curl https://dlang.org/install.sh | bash -s get-path --install
+/home/user/dlang/dmd-2.093.0/linux/bin64/dmd
+)
+
+$(CONSOLE ~/dlang/install.sh get-path ldc-1.23.0 --install
+/home/user/dlang/ldc-1.23.0/bin/ldc2
+)
+
+$(CONSOLE ~/dlang/install.sh get-path --dmd ldc --install
+/home/user/dlang/ldc-1.23.0/bin/ldmd2
+)
+
+$(CONSOLE ~/dlang/install.sh get-path --dub ldc --install
+/home/user/dlang/ldc-1.23.0/bin/dub
+)
+
+$(CONSOLE ~/dlang/install.sh get-path --dub dub-1.22.0 --install
+/home/user/dlang/dub-1.22.0/dub
+)
+
+$(CONSOLE ~/dlang/install.sh get-path --dub ldc,dub-1.22.0 --install
+/home/user/dlang/dub-1.22.0/dub
+)
+
+If a compiler installation isn't found and `--install` has not been passed, an error will be issued.
--- a/script/install.sh
+++ b/script/install.sh
@@ -172,6 +172,8 @@ display_path() {
 COMMAND=
 COMPILER=dmd
 VERBOSITY=1
+GET_PATH_AUTO_INSTALL=0
+GET_PATH_COMPILER=dc
 # Set a default install path depending on the POSIX/Windows environment.
 if posix_terminal; then
     ROOT=~/dlang
@@ -236,6 +238,7 @@ Commands
   uninstall     Remove an installed D compiler
   list          List all installed D compilers
   update        Update this dlang script
+  get-path      Path of the installed compiler executable
 
 Options
 
@@ -331,6 +334,36 @@ Description
 
   Update the dlang installer itself and the keyring.
 '
+            ;;
+
+        get-path)
+            log 'Usage
+
+  install.sh get-path <compiler>
+
+Description
+
+  Find the path of an installed D compiler.
+
+Options
+
+  --install         Install the compiler if it is not installed
+  --dmd             Find the DMD-alike interface instead
+  --dub             Find the DUB instance
+
+Examples
+
+  install.sh
+  install.sh get-path dmd
+  install.sh get-path dmd-2.093.0 --install
+  install.sh get-path --dmd ldc-1.19.0
+  install.sh get-path --dub ldc-1.19.0
+  install.sh get-path ldc-1.19.0 --install
+  install.sh get-path --dub dub-1.21.0
+'
+            log "$_compiler"
+            ;;
+
     esac
 }
 
@@ -338,7 +371,9 @@ Description
 
 parse_args() {
     local _help=
-    local _activate=
+    local _installAction=
+    local _getPathAction=
+    local _autoInstallFlag=
 
     while [[ $# -gt 0 ]]; do
         case "$1" in
@@ -359,10 +394,35 @@ parse_args() {
                 ;;
 
             -a | --activate)
-                _activate=1
+                if [ -n "$_installAction" ]; then
+                    fatal "$1 conflicts with --${_installAction}"
+                fi
+                _installAction="activate"
+                ;;
+
+            --dmd)
+                if [ -n "$_getPathAction" ]; then
+                    fatal "$1 conflicts with --${_getPathAction}"
+                fi
+                _getPathAction="dmd"
+                ;;
+
+            --dub)
+                if [ -n "$_getPathAction" ]; then
+                    fatal "$1 conflicts with --${_getPathAction}"
+                fi
+                _getPathAction="dub"
+                ;;
+
+            --install)
+                _autoInstallFlag=1
                 ;;
 
             use | install | uninstall | list | update)
+                COMMAND=$1
+                ;;
+
+            get-path)
                 COMMAND=$1
                 ;;
 
@@ -390,14 +450,41 @@ parse_args() {
         command_help $COMMAND
         exit 0
     fi
-    if [ -n "$_activate" ]; then
-       if [ "${COMMAND:-install}" == "install" ]; then
-           VERBOSITY=0
-       else
-           command_help $COMMAND
-           exit 1
-       fi
+    local command_="${COMMAND:-install}"
+    if [ -n "$_installAction" ] ; then
+        if [ "$command_" == "install" ]; then
+            VERBOSITY=0
+        else
+            logE "ERROR: --activate is not allowed for ${command_}."
+            command_help $COMMAND
+            exit 1
+        fi
     fi
+    if [ -n "$_autoInstallFlag" ] ; then
+        if [ "$command_" == "get-path" ]; then
+            GET_PATH_AUTO_INSTALL=1
+        else
+            logE "ERROR: --install is not allowed for ${command_}."
+            command_help $COMMAND
+            exit 1
+        fi
+    fi
+    if [ -n "$_getPathAction" ] ; then
+        if [ "$command_" == "get-path" ]; then
+            case "$_getPathAction" in
+                dmd|dub)
+                    GET_PATH_COMPILER=$_getPathAction
+                    ;;
+                *)
+                    fatal "Internal Error. Invalid get-path: $_getPathAction"
+            esac
+        else
+            logE "ERROR: --${_getPathAction} is not allowed for ${command_}."
+            command_help $COMMAND
+            exit 1
+        fi
+    fi
+
     IFS="," read -r _compiler _dub <<< "$COMPILER"
     if [ -n "${_dub:-}" ] ; then
         COMPILER="$_compiler"
@@ -411,51 +498,7 @@ parse_args() {
 run_command() {
     case $1 in
         install)
-            check_tools curl
-            if [ ! -f "$ROOT/install.sh" ] || [ ! -f "$ROOT/d-keyring.gpg" ] ; then
-                install_dlang_installer
-            fi
-            if [ -z "${2:-}" ]; then
-                fatal "Missing compiler argument for $1 command.";
-            fi
-            if [ -d "$ROOT/$2" ]; then
-                log "$2 already installed";
-            else
-                install_compiler "$2"
-            fi
-
-            # Only try to install dub if it wasn't the main compiler
-            if ! [[ $COMPILER =~ ^dub ]] ; then
-                if [ -n "${DUB:-}" ] ; then
-                    # A dub version was explicitly specified with ,dub-1.8.0
-                    DUB_BIN_PATH="${ROOT}/${DUB}"
-                    # Check whether the latest dub version needs to be resolved
-                    if ! [[ $DUB =~ ^dub- ]] ; then
-                        resolve_latest "$DUB"
-                        install_dub "dub-$DUB_VERSION"
-                    else
-                        install_dub "$DUB"
-                    fi
-                else
-                    # compiler was installed without requesting dub
-                    local -r binpath=$(binpath_for_compiler "$2")
-                    if [ -f "$ROOT/$2/$binpath/dub" ]; then
-                        if [[ $("$ROOT/$2/$binpath/dub" --version) =~ ([0-9]+\.[0-9]+\.[0-9]+(-[^, ]+)?) ]]; then
-                            log "Using dub ${BASH_REMATCH[1]} shipped with $2"
-                        else
-                            log "Using dub shipped with $2"
-                        fi
-                    else
-                        # no dub bundled - manually installing
-                        DUB_BIN_PATH="${ROOT}/dub"
-                        resolve_latest dub
-                        install_dub "dub-$DUB_VERSION"
-                    fi
-                fi
-            fi
-
-            write_env_vars "$2"
-
+            install_d "$1" "${2:-}"
             if posix_terminal; then
                 if [ "$(basename "$SHELL")" = fish ]; then
                     local suffix=.fish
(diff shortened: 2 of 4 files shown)


script/install.sh
```diff
@@ -150,6 +150,7 @@ Commands
   install       Install a D compiler (default command)
   uninstall     Remove an installed D compiler
   list          List all installed D compilers
+  where         Finds the binary for a D compiler (installs it if no compiler is found)
```
WebDrake: That seems an odd conflation of concerns.  Asking _where_ it is shouldn't auto-install it!

changelog/where.dd
```diff
@@ -0,0 +1,15 @@
...
+
+$(CONSOLE curl https://dlang.org/install.sh | bash -s where
+/home/seb/dlang/dmd-2.078.2/linux/bin64
+)
+
+$(CONSOLE ~/dlang/install.sh ldc where
+/home/seb/dlang/ldc-1.7.0/bin
+)
+
+As with `install`, if a compiler isn't found, it will be installed.
```
WebDrake: Why conflate these two concerns?  If one issues an instruction to `install`, then clearly one is asking for something to be installed if it isn't already.

If one is asking `where` something is then it seems reasonable that one might expect the response "nowhere", rather than have it be installed just so it can definitely be found.

script/install.sh
WebDrake: What motivation is there behind this assumption that `where` should install if not found?

WebDrake:
My questions seem to have got lost with the updated patch.  I'll summarize: why make `where` install a compiler if not found, rather than just report (in some appropriate way) that it is not found?

The simplest solution would probably be that `where` produces no output if the compiler is not installed, or the install location if it is.

Then, if in some script it's _wanted_ that a compiler be installed if not found, that could be implemented by querying `where` and issuing a separate `install` command if necessary.

wilzbach:
> My questions seem to have got lost with the updated patch. 

No worries. They weren't lost. I was about to reply to them.

> I'll summarize: why make where install a compiler if not found, rather than just report (in some appropriate way) that it is not found?

The updated patch removed `where` as a separate option. It now behaves analog to `-a` | `--activate` which is only an option for `install` (the default command) and will print the path to the sourceable `activate` script.

I hppe this answers all your questions as `--where` can't be called with anything else than the `install` command?

WebDrake:
I'm not sure it really feels like the right answer.  Could we take a pause on this and revisit tomorrow?  I need some sleep ;-)

I don't think it really satisfies what I see as _my_ use-case.  I would say that there are 2 different options that would satisfy what I'm looking for:

* provide an alternative flag for `install` that allows the installed compiler to be added to the `PATH` without overwriting all the other environment vars like `DC`, `DMD`, etc.  This feels a bit finnicky, though, because it could get tricky to track everything that's been added to the path (or not).

* provide an easy way to query the location of a named installed compiler, that would give the location if it's installed, or no output if not.

Bottom line, I _like_ the idea of the `where` command.  I just think it would satisfy my needs better if it just behaved as per the second bullet point above.  Is there a reason why that's difficult or problematic?

wilzbach:
> I'm not sure it really feels like the right answer. Could we take a pause on this and revisit tomorrow? I need some sleep ;-)

No worries. There's quite a bottleneck for `installer` (it's only @MartinNowak who can effectively approve PRs), so the review backlog is quite long - https://github.com/dlang/installer/pulls

> I don't think it really satisfies what I see as my use-case.

I'm not sure what's your use case, but it satisfies the need for `dlang/dmd` and `dlang/tools`.

-------------------

### DMD:

Previously:

```
source ~/dlang/install.sh "$DMD"
HOST_DMD=$DMD"
deactivate
```

Now

```
HOST_DMD=$(~/dlang/install.sh --where "$DMD")
```


--- 

### tools

```
~/dlang/install.sh install ldc
LDMD2=$(find ~/dlang -type f -name "ldmd2")
```

now:

```
LDMD2=$(~/dlang/install.sh install --where ldc)
```

> provide an alternative flag for install that allows the installed compiler to be added to the PATH without overwriting all the other environment vars like DC, DMD, etc. This feels a bit finnicky, though, because it could get tricky to track everything that's been added to the path (or not).

Not a huge fan of this. It's also more complicated than this PR.

> provide an easy way to query the location of a named installed compiler, that would give the location if it's installed, or no output if not.

How would that work with `~/dlang/install install ldc`?
I mean sure you could write a `/compiler_path` file in the root directory of the ldc archive and then do a `dirname` on the output of `--activate`, but then we would break code if we ever change the location of the `activate` script.

joseph-wakeling-sociomantic:
On thinking it over: yes, as a flag to `install` this is probably the best way of doing things.  It means that the answer is only possible at the time of install, and that's a useful simplification, because it means that there's no need to keep state hanging around, or rely on assumed rules about where compilers get placed.

A separate `where` command could work on the grounds that the install location for a given specified compiler is 100% deterministic (for the install script) based on the compiler name.  But of course this is not quite true when one considers generic `ldc`, `dmd`, `gdc` as opposed to specified versions.  So best avoided.

Something that's not clear from your above examples: would the result be the _directory_ where the compiler lives, or the full path to the compiler executable?  I ask because:

```
LDMD2=$(find ~/dlang -type f -name "ldmd2")
```

versus

```
LDMD2=$(~/dlang/install.sh install --where ldc)
```

... do not seem implicitly to be the same thing.  After all, with LDC and GDC there are _two_ compiler executables installed, with 2 different names (`ldc2` and `ldmd2`, and `gdc` and `gdmd`).

script/install.sh
```diff
@@ -192,6 +193,7 @@ Description
 Options
 
   -a --activate     Only print the path to the activate script
+  -w --where        Only print the path to the compiler binary
```
WebDrake: If I understand this code correctly, it's not the path to the compiler binary.  It's the path to the directory where the compiler binary has been installed.

script/install.sh
```diff
@@ -313,43 +328,25 @@ parse_args() {
...
+                    local suffix=.fish
+                fi
+                if [ "$VERBOSITY" -eq 0 ]; then
+                    echo "$ROOT/$2/activate${suffix:-}"
                 else
-                    log "Using dub shipped with $2"
+                    log "
+Run \`source     $ROOT/$2/activate${suffix:-}\` in your shell to use $2.
+This will se    tup PATH, LIBRARY_PATH, LD_LIBRARY_PATH, DMD, DC, and PS1.
+Run \`deacti    vate\` later on to restore your environment."
```
WebDrake: What's the reason for the weird whitespace on these lines?

changelog/where.dd
```diff
@@ -0,0 +1,15 @@
+`~/dlang/install.sh` now supports `--where`
+
+`--where` has been added as a new option to the install script.
+It offers a convenient way to get the path of a specific compiler.
```
WebDrake: `path of a specific compiler` is a bit vague -- `--where` doesn't provide the path to the compiler, it provides the directory where the compiler has been installed.

wilzbach:
> Something that's not clear from your above examples: would the result be the directory where the compiler lives, or the full path to the compiler executable? I ask because:
> ...
> ... do not seem implicitly to be the same thing. After all, with LDC and GDC there are two compiler executables installed, with 2 different names (ldc2 and ldmd2, and gdc and gdmd).

Contrary to the previous changelog entry, they were doing the same thing (`--where` was returning `<path-to-ldc-bin>/ldmd2`). Sorry for the confusion!

However, after thinking about this, I think I found an even better solution with `--dmd` and `--dc`.
This is analog to the existing `$DMD` and `$DC` variables that _this_ install script exposes.
I like this more because it's explicit what interface the user wants and no `dirname` hacks need to be done to switch the interface of the compiler from `ldmd` to `ldc`.

changelog/install-dmd-dc.dd
```diff
@@ -0,0 +1,27 @@
+`~/dlang/install.sh` now supports `--dmd` and `--dc`
+
+`--dmd` and `--dc` has been added as a new option to the install script.
+These flags work analogous to the existing `--activate` flag and
+print information about the installed compiler binary:
+
+$(UL
+    $(LI `--dmd` returns the path of a specific compiler executable (DMD-like interface))
+    $(LI `--dc` returns the path of a specific compiler executable (custom compiler interface))
```
joseph-wakeling-sociomantic: Can this be the same as what `--dmd` gives?  Might be worth noting if so.

changelog/install-dmd-dc.dd
```diff
@@ -0,0 +1,27 @@
...
+)
+
+$(H4 Examples)
+
+$(CONSOLE curl https://dlang.org/install.sh | bash -s --dmd
+/home/seb/dlang/dmd-2.078.2/linux/bin64/dmd
```
joseph-wakeling-sociomantic: What would `--dc` give you if the compiler is `dmd`?

changelog/install-dmd-dc.dd
```diff
@@ -0,0 +1,27 @@
...
+print information about the installed compiler binary:
+
+$(UL
+    $(LI `--dmd` returns the path of a specific compiler executable (DMD-like interface))
+    $(LI `--dc` returns the path of a specific compiler executable (custom compiler interface))
+)
+
+$(H4 Examples)
+
+$(CONSOLE curl https://dlang.org/install.sh | bash -s --dmd
```
joseph-wakeling-sociomantic: Not sure it's really helpful to put in this `curl` stuff.  It distracts from the example of how to use the `install.sh` script itself.

changelog/install-dmd-dc.dd
```diff
@@ -0,0 +1,27 @@
+`~/dlang/install.sh` now supports `--dmd` and `--dc`
+
+`--dmd` and `--dc` has been added as a new option to the install script.
+These flags work analogous to the existing `--activate` flag and
+print information about the installed compiler binary:
+
+$(UL
+    $(LI `--dmd` returns the path of a specific compiler executable (DMD-like interface))
```
joseph-wakeling-sociomantic: What will this be if the compiler does not expose a DMD-like interface?

changelog/install-dmd-dc.dd
wilzbach: It's how you get the installer script and would use it for a CI script.
It also provides a seamless experience for readers who want to try out the new flags as they can just copy/paste the command.

changelog/install-dmd-dc.dd
wilzbach: The same (see the tests). The changelog isn't intended to be a full documentation, but to raise awareness of new features or improvements. dlang.org/install.html will come with more extensive documentation.

changelog/install-dmd-dc.dd
wilzbach: I don't think SDC will grow into a full compiler any time soon and even if, for a seamless experience it most likely will implement the DMD interface (IIRC it already does so).
The same applies for other compilers and if there's ever a case in a far far far distant future where we would want to include a new compiler, but are too lazy to wrap the DMD interface around it, `fatal` error is the Unix way.
Also as mentioned this is the changelog entry - not the documentation, so imho there's no need to go into details here.

WebDrake: > Also as mentioned this is the changelog entry - not the documentation, so imho there's no need to go into details here.

I didn't ask for the changelog to be altered.  I asked what the result will be if the compiler does not expose a DMD-like interface.

changelog/install-dmd-dc.dd
WebDrake: Sure, but it would be clearer if the example was annotated to that effect, e.g. `download and run install script, printing out the installed compiler`.

script/install.sh
```diff
@@ -192,6 +195,8 @@ Description
 Options
 
   -a --activate     Only print the path to the activate script
+  --dmd             Only print the path to the compiler binary (e.g. dmd, ldmd2, gdmd)
+  --dc              Only print the path to the compiler binary (e.g. dmd, ldc, gdc)
```
WebDrake: * What will happen if I use both `--dmd` and `--dc` together?

* It looks like this documentation could be clearer about the distinction between the flags, i.e. that one is for the DMD-alike compiler binary

MartinNowak:
> HOST_DMD=$(~/dlang/install.sh --where "$DMD")

Sorry, I've only grokked part of the discussion.

Just using the compiler without the environment is not reliable, as the sc.ini/dmd.conf in your user home take precendence over the ones next to the compiler binary :/.
I'm still convinced that a lot of non-sense is rooted on this single problem. Why would a makefile need an absolute path to the compiler? Just so that it can for usage of `-conf=` or `-conf=path/to/dmd.conf`.
https://github.com/dlang/dmd/pull/4256

wilzbach:
> Just using the compiler without the environment is not reliable, as the sc.ini/dmd.conf in your user home take precendence over the ones next to the compiler binary :/.
> I'm still convinced that a lot of non-sense is rooted on this single problem. 

Urgh I wasn't even aware of this. Let's fix this mess then -> https://github.com/dlang/dmd/pull/7915

> Sorry, I've only grokked part of the discussion.

The idea is to have a `--dmd` and `--dc` flag which returns the location of the compiler binary (they return the same output as `$DMD` and `$DC`), s.t
- the environment doesn't need to be loaded which turns out is very error-prone (e.g. our recent CI fiasco at dmd)
- multiple compiler can be installed and used in parallel without needing to switch environments (in @WebDrake's use case for dlang/tools the rdmd testsuite even accepts multiple compiler binaries)

> Just using the compiler without the environment is not reliable, as the sc.ini/dmd.conf in your user home take precendence over the ones next to the compiler binary :

Well, that wouldn't be a huge problem for CIs (the main motivation for this feature), but of course one could always do sth. like this with `dirname`:

```shell
DMD=$(~/dlang/install.sh 2.078.3  --dmd)
make DMD="$DMD" DFLAGS="-conf="$(dirname "$DMD")/dmd.conf)"
```

In any case the motivation was that `--dmd` is cleaner and less prone to errors than `find ~/dlang -type f -name "ldmd2"` [which is currently used](https://github.com/dlang/tools/blob/master/travis.sh).

WebDrake:
@MartinNowak a specific example where multiple compilers are wanted comes from here:
https://github.com/dlang/tools/blob/master/travis.sh#L5-L16

The goal there is to have one build of `rdmd` tested with multiple different compilers passed via the `--compiler` flag.

Now that said, I'm not wedded to that particular way of doing things: it was a step forward given the existing test setup, but there are multiple things that probably can and should be changed.

Note that the workaround I implemented for that use case (i.e. using `find`) seems to be working OK for the time being, so from my side there is no urgency to this `--where` flag or anything like it.

changelog/where.dd
CyberShadow: I think installation should be available as a flag for all commands which imply using a specific installer version.

I.e. `source ~/dlang/install.sh activate -i 2.080.0` should install 2.080.0 if it's not installed, then activate it.

changelog/install-dmd-dc.dd
```diff
@@ -0,0 +1,27 @@
+`~/dlang/install.sh` now supports `--dmd` and `--dc`
```
CyberShadow: These are not options, they're actions. They should be named "get-dmd-path" and "get-dc-path" OSLT.

Am I missing something?

script/install.sh
CyberShadow: Yep, conceptually these are actions, not flags.

AndrewEdwards:
@wilzbach @WebDrake @joseph-wakeling-sociomantic @CyberShadow @MartinNowak 

What is the current stand on this? Is there actually something preventing this from being merged? With the exception of the requested change from "--dmd" and "--dc" to "get-dmd-path" and "get-dc-path" all other concerns appears to have been addressed. @wilzbach can you address @CyberShadow's concern and resolve the conflicts to prepare this for merger? @MartinNowak, what other concerns if any do you have about this pull request?

changelog/install-dmd-dc.dd
wilzbach: This can't happen as `install.sh` only supports GDC, DMD and SDC. If there's ever a new compiler which would be added to `install.sh`, it would fail, I presume.

changelog/install-dmd-dc.dd
wilzbach: I added an example.

changelog/install-dmd-dc.dd
wilzbach: > What would --dc give you if the compiler is dmd?

I added an example + it's not the documentation, just a changelog.

wilzbach:
> @wilzbach can you address @CyberShadow's concern and resolve the conflicts to prepare this for merger? 

Ok. I reworked it. I combined them under a single `get-path` action and added support for getting the path to the requested dub binary as well.

CyberShadow:
Is it possible to remove those annoying CodeCov annotations? They make the diff unreadable.

AndrewEdwards:
I see what you mean. Sorry, not sure how to do that. I'll defer to @wilzbach.

CyberShadow:
```js
for (var n of document.querySelectorAll('.check-annotation')) n.parentElement.remove(n);
```
in the console does it.

test/common.sh
```diff
@@ -4,3 +4,19 @@ set -eu -o pipefail
...
 INSTALLER="$ROOT/script/install.sh"
+
+assert() {
+    actual="$1"
+    expected="$2"
+    if [ "$actual" != "$expected" ] ; then
+        echo "Actual: $actual"
+        echo "Expected: $expected"
+        exit 1
+    fi
```
CyberShadow, line 15: For this I would simply write `diff -u <(cat <<< "$1") <(cat <<< "$2")`, but this is OK too.

AndrewEdwards:
I'll need to set up my environment first. Currently just using the web interface. As a temporary fix, you can disable them per file by unchecking the show annotations option.

disable-annotations

CyberShadow:
Ah, thanks. I missed that. That's much simpler.

CyberShadow (approved):
Thank you. Good work on the thorough test script.

wilzbach:
Wow. I didn't think this would ever get merged. Thanks!

I opened a documentation PR -> https://github.com/dlang/dlang.org/pull/2849

PR #298, opened 2018-02-17
temporarily remove chmgen (d.chm, dman) from release

MartinNowak:
- Chmgen and the d.chm Windows file are unreasonably unreliable
  there were 4 issues responsible for 39 failing nightly builds in
  2016, which are very time-consuming to debug and fix.
  It seems that the basic approach to parse the html docs into
  structured information is flawed and not sustainable.
- chm files are nice for offline usage, but it's a very minor tool
  on which we spent a very unproportionate amount of resources.
- dman doesn't even offer offline capabilities, and can be easily
  replaced by a browser shortcut to dlang.org's search field.

Also https://github.com/dlang/dlang.org/pull/2212 and https://forum.dlang.org/post/alcerzuvydfutsiuelsk@forum.dlang.org.

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -512,21 +512,8 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
             {
                 changeDir(cloneDir~"/dlang.org");
                 run(makecmd~" DOC_OUTPUT_DIR="~origDir~"/docs release");
-                // put into docs/ folder which gets copied to all other platforms
-                copyFile("d-tags-release.json", origDir~"/docs/d-tags-release.json");
                 // copy generated man pages to docs/man which gets copied to all other platforms
                 copyDir(cloneDir~"/dmd/generated/docs/man", origDir~"/docs/man");
-            }
-        }
-        else version (Windows)
-        {
-            if (bits == Bits.bits32)
-            {
-                // copy linux-generated docs into dlang.org
-                copyDir(origDir~"/docs", cloneDir~"/dlang.org");
-                // and build the d.chm file
-                changeDir(cloneDir~"/dlang.org");
-                run(makecmd~" chm");
             }
         }
     }
@@ -538,12 +525,6 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
         run(makecmd~" rdmd");
         run(makecmd~" ddemangle");
         run(makecmd~" dustmite");
-        if (!skipDocs)
-        {
-            // use tags built with linux docs
-            copyFile(origDir~"/docs/d-tags-release.json", cloneDir~"/tools/d-tags.json");
-            run(makecmd~" dman");
-        }
 
         removeFiles(cloneDir~"/tools", "*.{"~obj~"}", SpanMode.depth);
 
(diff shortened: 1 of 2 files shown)


MartinNowak:
There was a previous discussion ([Anybody still using the chm docs](https://forum.dlang.org/post/cuxcxgwhzonctgnizqwo@forum.dlang.org)) and searching for [chmgen](https://forum.dlang.org/search?q=chmgen) should give you some impression on the number of issues.
If someone still wants to maintain this, please provide this as an external service (download) that doesn't interfere with releases.

CyberShadow:
I am maintaining chmgen. Feel free to contact me about it directly. So far, I have not been aware of any critical issues caused by it recently.

> Chmgen and the d.chm Windows file are unreasonably unreliable
> there were 4 issues responsible for 39 failing nightly builds in
> 2016, which are very time-consuming to debug and fix.

I was not aware of this. Still, that sounds like a failure of CI. Please provide some examples, so that we can improve the CI.

> and searching for chmgen should give you some impression on the number of issues.

Those search results are not relevant.

We need chmgen to lint the generated HTML, so we should not remove it.

MartinNowak:
> I am maintaining chmgen. Feel free to contact me about it directly. So far, I have not been aware of any critical issues caused by it recently.

Sorry, d.chm builds are untested and regularly breaks without anyone noticing, so effectively I'm maintaining it in order to build releases.
For sure we can keep the tool somewhere and maybe even the tool, but the approach is whacky and I'm no longer willing to accept this as a release obstacle or spent more time fixing obscure failures.

MartinNowak:
> We need chmgen to lint the generated HTML, so we should not remove it.

OK, but we could still remove d.chm and dman from the release.

MartinNowak:
Merging this for now so I can build the 2.079 beta (https://github.com/dlang/dlang.org/pull/2213#issuecomment-366454866).

WalterBright:
What was the rationale for removing dman? I use it all the time. And no, I don't have dlang.org open all the time.

p0nce:
Chiming in as a data point: I've never heard anyone use either of the CHM (or dman fwiw). When offline I read phobos/druntime sources before anything else.

MartinNowak:
> What was the rationale for removing dman? I use it all the time. And no, I don't have dlang.org open all the time.

This is not your old dman which only knew a few tags.
It's based on the same complex mechanism that uses ddoc macros to generate json, which are then parsed by a processor (chmgen) to get tags. This is combined with extracting links from html files to generate the chm file (might not be necessary for dman). So in total complex enough that I don't have the capacity to get into this. It does occasionally fail though, and when it does, releases are blocked and I cannot do anything.
There doesn't seem any point in including rarely used tools directly in the release, those could easily be build and distributed separately, e.g. using one of the CI systems.

wilzbach:
> There doesn't seem any point in including rarely used tools directly in the release, those could easily be build and distributed separately, e.g. using one of the CI systems.

Especially given that many replacements that work offline exists, e.g. http://devdocs.io/d

CyberShadow:
Instead of posting the same thing in multiple places, it's better to merge related discussions. Let's continue in the dlang.org issue: https://github.com/dlang/dlang.org/pull/2213#issuecomment-368263622

PR #301, opened 2018-02-21
Allow installing dub in standalone

wilzbach:
This allows using the installer script to install an individual version of DUB.

The main motivation for this are the failing DUB + Vibe.d Travis builds which still test very old DMD versions which don't come with the latest DUB binaries.

In the future we might support something like `dmd-2.072.0+dub` (latest dub) or `dmd-2.072.2+dub.1.7.2`, but for now this allows @s-ludwig to do add this extra line to update to the latest dub:

```
. $(~/dlang/install.sh dub -a)
```


script/install.sh
```diff
@@ -354,16 +355,19 @@ run_command() {
...
+                    if [[ $("$ROOT/$2/$binpath/dub" --version) =~ ([0-9]+\.[0-9]+\.[0-9]+(-[^, ]+)?) ]]; then
+                        log "Using dub ${BASH_REMATCH[1]} shipped with $2"
+                    else
+                        log "Using dub shipped with $2"
+                    fi
                 else
-                    log "Using dub shipped with $2"
+                    DUB_BIN_PATH="${ROOT}/dub"
+                    install_dub
                 fi
```
wilzbach: I deliberately avoided updating the default `dub` symlink (which happens in `install_dub`), though there are pros and cons for this.

script/install.sh
```diff
@@ -458,11 +463,31 @@ resolve_latest() {
...
+        dub)
+            local mirrors=(
+                "http://code.dlang.org/download/LATEST"
+                "http://dlang.github.io/dub/LATEST"
+            )
+            logV "Determining latest dub version (${mirrors[0]})."
+            DUB_VERSION="$(fetch "${mirrors[@]}")"
+            local DUB="dub-${DUB_VERSION}"
+            if [ -z "$1" ] ; then
+                COMPILER="$DUB"
```
wilzbach: I wasn't sure on the best way to deal with this.
The problem is that I wanted to reuse the logic here in `install_dub`, but in `install dmd` path, we can't set `COMPILER`, but for `install dub` we should.

The missing coverage is because currently we only use specific versions on for the CI.

script/install.sh
```diff
@@ -542,6 +567,17 @@ install_compiler() {
...
 
+    elif [[ $1 =~ ^dub-([0-9]+\.[0-9]+(\.[0-9]+)?(-.*)?)$ ]]; then
+        local ver=${BASH_REMATCH[1]}
+        local url="https://github.com/ldc-developers/ldc/releases/download/v$ver/ldc2-$ver-$OS-$ARCH.tar.xz"
+        if [ $OS != linux ] && [ $OS != osx ]; then
+            fatal "no ldc binaries available for $OS"
+        fi
+        local mirrors=(
+            "http://code.dlang.org/files/dub-$ver-$OS-$ARCH.tar.gz"
+            "https://github.com/dlang/dub/releases/download/$ver/dub-$DUB_VERSION-$OS-$ARCH.tar.gz"
```
wilzbach: See also: https://github.com/dlang/installer/pull/273

script/install.sh
```diff
@@ -580,6 +616,8 @@ download_and_unpack() {
     download "$tmp/$name" "$do_verify" "${urls[@]}"
     if [[ $name =~ \.tar\.xz$ ]]; then
         tar --strip-components=1 -C "$tmp" -Jxf "$tmp/$name"
+    elif [[ $name =~ \.tar\.gz$ ]]; then
+        tar -C "$tmp" -xf "$tmp/$name"
```
wilzbach: The dub binary is in the root-level.

script/install.sh
```diff
@@ -667,67 +708,120 @@ write_env_vars() {
...
+            local dmd=
+            ;;
     esac
 
     logV "Writing environment variables to $ROOT/$1/activate"
     cat > "$ROOT/$1/activate" <<EOF
 deactivate() {
     export PATH="\$_OLD_D_PATH"
+EOF
+    if [ -n "$libpath" ] ; then
```
wilzbach: The following code writes thin `activate` scripts that only set the variables that are actually set.
In other words things like `LIBRARY_PATH` aren't written for `dub`

test/all.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+. $DIR/common.sh
+
+for file in $(find "$DIR" -name "*.sh") ; do
```
wilzbach: Motivation: it's a lot more convenient to have small files as they can be run individually.
I have already run into this before, e.g. https://github.com/dlang/installer/pull/297 (contains the same idea with `test/all.sh`)

wilzbach:
FYI this has now been used by Vibe.d (https://github.com/vibe-d/vibe.d/pull/2140) and Dub (https://github.com/vibe-d/vibe.d/pull/2140) for about a month with the builds being a lot more greenish...

wilzbach:
At DConf Martin was a bit sceptical about using `+` as it's a valid SemVer character.
How about using `@`, `|` or `:` instead?

script/install.sh
```diff
@@ -389,6 +408,10 @@ Run \`deactivate\` later on to restore your environment."
                 fatal "Missing compiler argument for $1 command.";
             fi
             uninstall_compiler "$2"
+            if [ -n "${DUB+x}" ] ; then
+                echo $DUB
```
CyberShadow: Debug code?

script/install.sh
```diff
@@ -454,11 +478,11 @@ resolve_latest() {
...
-            if [[ ! $COMPILER =~ -[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]] &&
-               [[ ! $COMPILER =~ -[0-9][.][0-9]{3}[.][0-9]{1,3}(-[0-9]{1,3})? ]]
+            if [[ ! $compiler =~ -[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]] &&
+               [[ ! $compiler =~ -[0-9][.][0-9]{3}[.][0-9]{1,3}(-[0-9]{1,3})? ]]
             then
-                local url=http://nightlies.dlang.org/$COMPILER/LATEST
-                logV "Determing latest $COMPILER version ($url)."
+                local url=http://nightlies.dlang.org/$compiler/LATEST
+                logV "Determing latest $compiler version ($url)."
                 COMPILER="dmd-$(fetch "$url")"
```
CyberShadow: Variables differing only by case? Is that necessary? :)

script/install.sh
```diff
@@ -673,67 +729,120 @@ write_env_vars() {
...
-    set -e _OLD_D_LD_LIBRARY_PATH
+EOF
+
+if [ -n "$dmd" ] ; then
+    cat >> "$ROOT/$1/activate.fish" <<EOF
     set -e DMD
     set -e DC
+EOF
+fi
+cat >> "$ROOT/$1/activate.fish" <<EOF
```
CyberShadow: Might be less messy to have one scope per file instead of all these appends.

https://github.com/koalaman/shellcheck/wiki/SC2129

test/all.sh
```diff
@@ -0,0 +1,12 @@
...
+    if ! ( [[ "$file" == */all.sh ]] || [[ "$file" == */common.sh ]] ) ; then
```
CyberShadow:
```bash
     if [[ "$file" != */all.sh && "$file" != */common.sh ]] ; then 
```

But really, just put the actual test files in their own directory, or give them a distinguished name (like `test-*.sh`).

https://github.com/CyberShadow/aconfmgr/tree/master/test/t

wilzbach:
> How about simply `dmd-2.072.2,dub.1.7.2`?

Ok. Went with this ;-)

wilzbach:
> 24 days ago

Any objections?

CC @ZombineDev @thewilsonator

thewilsonator:
Not really, but I don't understand the code very well.

PR #302, opened 2018-02-21
Always update the install script

wilzbach:
If we go with this, imho we should reopen https://github.com/dlang/installer/pull/274 (config file) and allow people to opt-out of this.
There are certainly use cases in which always updating the installer can be annoying, especially if you use the install script with an already installed compiler this will add a few seconds extra to every activation.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -342,9 +342,7 @@ run_command() {
     case $1 in
         install)
             check_tools curl
-            if [ ! -f "$ROOT/install.sh" ]; then
-                install_dlang_installer
-            fi
+            install_dlang_installer || log "Updating the installer failed."
             if [ -z "${2:-}" ]; then
                 fatal "Missing compiler argument for $1 command.";
             fi


wilzbach:
> I'd be in favor of some regular (weekly) etag based check.

Where can we store the etag?
Something like `${XDG_CACHE_HOME:-$HOME/.cache}/dlang-installer/self-etag` ?

> This seems a bit too blunt and also unexpected for users. 

Fair enough.

CyberShadow:
I think this is fine to do when installing a new version.

Currently "install" (with a specific version) does nothing if that version compiler is already installed, and I think that should continue not performing any network activity.

You could also make use of the `Last-Modified` and `If-Modified-Since` headers to avoid downloading/installing it if it's up to date. `wget` has this with its `-N` switch; for curl, parsing the headers will be necessary.

AndrewEdwards:
@wilzbach, is it safe to assume that you no longer plan on pursuing this further?

wilzbach:
> @wilzbach, is it safe to assume that you no longer plan on pursuing this further?

Well, I'm definitely not actively working on this. Sorry.
FYI: There was a recent attempt on this (https://github.com/dlang/installer/pull/457) that I unfortunately had to revert as it was merged without review.

PR #305, opened 2018-02-23
build LLD against static VC runtime library

rainers:
see https://issues.dlang.org/show_bug.cgi?id=18509


rainers:
We could also switch to LLVM/LLD 6.0 which is about to be released.

MartinNowak:
> We could also switch to LLVM/LLD 6.0 which is about to be released.

You think it would solve any particular issue?

rainers:
> You think it would solve any particular issue?

No, but maybe some undiscovered ones. I tried LLD/LLVM master against https://issues.dlang.org/show_bug.cgi?id=18510, but it's still there. I've added an issue in the LLVM bugzilla, though.

PR #306, opened 2018-02-25
18522 - [REG 2.079-b2] MinGW import libraries interfere with platform SDK

rainers:
build mingw libraries into a mingw sub folder as expected by dmd


rainers:
Unfornately SourceForge is getting unusable slow, should we put the mingw tarballs on downloads.dlang.org, too?

Alternatively, there is https://github.com/mirror/mingw-w64 which seems a bit more up to date, but also has a different source organization, i.e. more work to adapt.

rainers:
Finally built successfully (after removing the sourceforge redirection). @MartinNowak please update http://downloads.dlang.org/other/mingw-libs-5.0.2.zip from https://ci.appveyor.com/project/4wil/installer/build/artifacts

MartinNowak:
BTW @rainers, please use milestones to mark PRs that need to merged before a release so I can find them in time.

rainers:
> BTW @rainers, please use milestones to mark PRs that need to merged before a release so I can find them in time.

Ok, will try to remember next time. I thought targeting stable is enough, though that might not be easily searchable. It wouldn't have been appropriate for this PR anyway.

PR #309, opened 2018-03-13
Remove old licenses from the OSX description

wilzbach:
Everything is now Boost-licensed and also I don't think there's a need to ship the default dummy license.

(untested)


wilzbach:
CC @jacob-carlborg do you still package this?

jacob-carlborg:
> CC @jacob-carlborg do you still package this?

Not sure what you’re referring to but the installer for macOS is still built and the first choice on dlang.org.

jacob-carlborg:
But as far as the license it looks good.

jacob-carlborg:
Perhaps the mentioned macOS version at the bottom should be updated.

wilzbach:
> Perhaps the mentioned macOS version at the bottom should be updated.

https://github.com/dlang/installer/pull/308 (that's how I discovered this XML file)

> Not sure what you’re referring to but the installer for macOS is still built and the first choice on dlang.org.

Line 47 in the diff:

```
Mac OS X DMD package by Jacob Carlborg <doob@me.com>
```

jacob-carlborg:
I initially created the package, or what’s needed to build the package. But I don’t build it, I assume it’s automated.

PR #310, opened 2018-03-24
Make the install script compatible with more shells

wilzbach:
Process substitution is availably only in bash/zsh, but not in general for sh.

https://superuser.com/questions/1059781/what-exactly-is-in-bash-and-in-zsh

This is an effort to allow using the installer on platforms on which bash isn't installed.
Also `curl ... | sh` would be nice to use eventually.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -2,6 +2,8 @@
 
 set -uexo pipefail
 
+echo $SHELL
+
 compilers=(
     dmd-2.069.2
     dmd-2.071.2
@@ -52,7 +54,7 @@ for idx in "${!compilers[@]}"
 do
     compiler="${compilers[$idx]}"
     echo "Testing: $compiler"
-    ./script/install.sh $compiler
+    $SHELL ./script/install.sh $compiler
 
     . ~/dlang/$compiler/activate
     compilerVersion=$($DC --version | sed -n 1p)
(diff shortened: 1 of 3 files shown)


wilzbach:
FYI: we currently support fish for the activation script, but that's failing at the moment, e.g.

https://travis-ci.org/dlang/installer/jobs/357737014

We should probably add this as another job if we want to continue supporting fish.

travis.sh
```diff
@@ -91,25 +93,25 @@ done
...
 for cmd in "${cmds[@]}"
 do
-    ./script/install.sh --help | grep -F "$cmd" >/dev/null
-    ./script/install.sh -h | grep -F "$cmd" >/dev/null
-    ./script/install.sh "$cmd" --help | tr -d '\n' | grep -q "Usage\s*install.sh $cmd" >/dev/null
-    ./script/install.sh "$cmd" -h | tr -d '\n' | grep -q "Usage\s*install.sh $cmd" >/dev/null
+    $SHELL /script/install.sh --help | grep -F "$cmd" >/dev/null
+    $SHELL /script/install.sh -h | grep -F "$cmd" >/dev/null
+    $SHELL /script/install.sh "$cmd" --help | tr -d '\n' | grep -q "Usage\s*install.sh $cmd" >/dev/null
+    $SHELL /script/install.sh "$cmd" -h | tr -d '\n' | grep -q "Usage\s*install.sh $cmd" >/dev/null
```
CyberShadow: This doesn't look right. Absolute paths?

AndrewEdwards:
@wilzbach Need you eyes on this one. Still working to make this happen?

PR #311, opened 2018-03-24
Issue 18658 - Build gnumake

wilzbach:
If I understand your branch concept correctly, this would need to target `build-gmake` (though I think it would be better to have all scripts in master and just selective enable/disable the appveyor hook on demand)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,15 @@
+os: Visual Studio 2017
+
+environment:
+  matrix:
+    - ARCH:                x86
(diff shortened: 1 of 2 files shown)


AndrewEdwards:
@thewilsonator Request review and approval.

wilzbach:
Luckily we don't need this anymore as DMD can be built 100% without any kind of make these days (:
Druntime and Phobos still have their old build files, but for them we don't need GNUMake on Windows.

PR #314, opened 2018-03-30
Fix Issue 18649 - curl on Ubuntu 18.04 depends on libcurl4, .deb installer depends on libcurl3

wilzbach:
Here's my attempt at fixing this by making libcurl3 an optional dependency.


MartinNowak:
Looking at this, though it likely won't make it to 2.079.1 yet. Currently writing an installation test based on docker images.
Still working on reproducing the collision though, `apt-get --fix-broken install` (the step after `dpkg -i dmd_2.079.0-0_amd64.deb` just "removes" dmd without deinstalling it, thus afterwards the compiler appears to work, but in fact has been removed from the dependency tracking (or so).
I think we should use alternates `libcurl3 | libcurl4 | libcurl4-openssl`, also note that there is a [libcurl4](https://packages.ubuntu.com/bionic/libcurl4) in bionic (not in artful 17.x though).

jordisayol:
I'm agree with @MartinNowak. Keep "libcurl3" and add "libcurl4" as an "or" dependency.

MartinNowak:
I also want to have it in recommends (installed by default) as dub and std.net.curl need it.
Keeping libcurl3 is mostly for compat with older ubuntus and debians.
Still testing the approach, but targeted at 2.080.0.

jordisayol:
If you want to be installed it by default, use "Depends:" field, "Recommends:" can be disabled to install by default by the user.
You can also chose the preference on what the system will install first:

Depends: ... , libcurl4 | libcurl3, ....

MartinNowak:
> "Recommends:" can be disabled to install by default by the user

I actually wanted recommends, people disabling that should know what they do, and it allows to skip installation on systems that don't net it.
Turns out though that the `dpkg -i dmd_2.079.1-0_amd64.deb && apt-get install --fixup-broken` method does not install recommended packages, so I'll keep it in the depends section.

PR #315, opened 2018-03-31
remove duplicate curl --fail option

MartinNowak:


wilzbach:
> Wget has been designed for robustness over slow or unstable network connections; if a download fails due to a network problem, it will keep retrying until the whole file has been retrieved. If the server supports regetting, it will instruct the server to continue the download from where it left off.

> ...

> ‘--tries=number’
> Set number of tries to number. Specify 0 or ‘inf’ for infinite retrying. The default is to retry 20 times, with the exception of fatal errors like “connection refused” or “not found” (404), which are not retried.

https://www.gnu.org/software/wget/manual/wget.html

MartinNowak:
> Sorry. BTW what do you think about using curl's built-in retry loop?

How would you retry a partial download against stdout? What's wrong with out current approach btw?
Yes wget is a better tool for downloads but seems to have a smaller installation base, and we'd still need to handle mirror support.
Overall doesn't the current approach seems to work, why change it?

PR #317, opened 2018-04-05
Add support for rdmd in gdc + set $RDMD variable

wilzbach:
- unifies scripting because now it can be assumed that `rdmd` exists and that `$RDMD` will point to it.


wilzbach:
Urgh. Some GDC versions have lower Phobos versions despite having a higher release number :/

wilzbach:
Also in relation with https://github.com/dlang/installer/pull/297 it might be interesting to add support for a `--rdmd` flag.

AndrewEdwards:
@wilzbach - Still moving forward? Please resolve conflicts.

PR #318, opened 2018-04-16
fix Issue 18649 - curl on Ubuntu 18.04 depends on libcurl4, ...

MartinNowak:
... .deb installer depends on libcurl3

- depend on libcurl4 | libcurl3 alternative
- degrading libcurl from dependency to recommends would be welcome,
  but does not work because recommendations are not installed for raw
  .deb packages (--fix-broken only installs dependencies).
  `dpkg -i dmd_2.079.1-0_amd64.deb && apt-get --fix-broken install`
- also add suggested libcurl-dev alternative
- also add install test script for supported debian and ubuntu releases

Supersedes #314

### DIFF
--- a/linux/dmd_deb.sh
+++ b/linux/dmd_deb.sh
@@ -344,8 +344,8 @@ else
 
 
 	# set deb package dependencies
-	DEPENDS="libc6, libc6-dev, gcc, libgcc1, libstdc++6, libcurl3"
-	SUGGESTS="libcurl4-openssl-dev, gcc-multilib"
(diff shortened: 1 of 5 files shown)


MartinNowak:
This is now integrated with the release.sh script, finally some basic binary testing before uploading.
https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0/0ccbd01925028a2d9d74e7263ca05a5cd5718e2e

wilzbach:
Really cool!
BTW any chance that you can move the gist into a repo, s.t. people can try to make PRs to it?

PR #323, opened 2018-05-08
Make install.sh group-writable

wilzbach:
This should allow the install.sh script to stay group-writable by the dlang.org group.


MartinNowak:
Let's rather delete and readd the file without g+w, so we can clearly see which user deployed the latest state. At lest on nightlies.dlang.org the containing folder has a setgid flag.

PR #330, opened 2018-07-03
Build the newest version of MinGW

wilzbach:
We currently use a more than one year old release: https://mingw-w64.org/doku.php/versions


wilzbach:
Sorry 5.0.2 seems to be the latest. I was confusing it with MinGW-64.
See also: https://sourceforge.net/projects/mingw/files/MinGW/Base/w32api/

PR #332, opened 2018-07-05
Issue 19051 - Undefined functions Set/GetWindowLongPtr in mingw libs

rainers:
these symbols do not exist for Win32 and are still missing in the 6.0 branch for mingw: https://sourceforge.net/p/mingw/mingw-org-wsl/ci/6.0-exp/tree/wslapi/lib/

mingw-w64 has them (https://github.com/mirror/mingw-w64/blob/master/mingw-w64-crt/lib-common/user32.def.in#L849) but basing the generation on a different distro is not (yet) reasonable.

### DIFF
--- a/windows/mingw/buildsdk.d
+++ b/windows/mingw/buildsdk.d
@@ -54,6 +54,20 @@ bool def2implib(bool x64, string f, string dir, string linkopt = null)
     if (pos < 0)
         return false;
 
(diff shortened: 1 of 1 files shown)


wilzbach:
To update the file in the master branch: https://github.com/dlang/installer/pull/333

PR #333, opened 2018-07-11
Update mingw to 5.0.2-1

wilzbach:
See https://github.com/dlang/installer/pull/332

I named the zip file mingw-libs-5.0.2-1.zip, uploaded it and its signature to S3.
It looks like that's all that's needed.


wilzbach:
Oh I didn't know how to update the index.html listing page, but I figured it out now:

rainers:
Ah, I thought that page would be automatically generated from a directory on the server.

LGTM, but I don't seem to have any merge rights here.

wilzbach:
> Ah, I thought that page would be automatically generated from a directory on the server.

I thought so too. For future references, one needs to run the scripts at https://github.com/dlang/downloads.dlang.org to generate the HTML index listings.

PR #334, opened 2018-07-31
fix uninitialised suffix on macOS

John-Colvin:
The problem was introduced in https://github.com/dlang/installer/pull/239 and presumable copied in https://github.com/dlang/installer/pull/218


John-Colvin:
You would think so, yes.

PR #335, opened 2018-07-31
less gnu-y alternative to find -printf

John-Colvin:
because not everywhere has gnu tools (e.g. macOS)

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -759,7 +759,7 @@ list_compilers() {
              -not -name install.sh \
              -not -name d-keyring.gpg \
              -not -name '.*' \
-             -printf "%f\n" | \
+             -exec basename {} \; | \
             grep . # fail if none found
     fi
 }


John-Colvin:
I would have expected this to be caught already by travis as well.

wilzbach:
> I would have expected this to be caught already by travis as well.

Same here. Though considering that this script is used by Travis for their builtin D support, I'm partially glad it did work.
I couldn't see anything in the logs of the master branch:

https://travis-ci.org/dlang/installer/jobs/399790376

Any idea why it's working fine on Travis OSX?

wilzbach:
It's not mentioned here: https://docs.travis-ci.com/user/reference/osx, but yeah apparently newer brew versions even have a special flag for this (https://apple.stackexchange.com/questions/69223/how-to-replace-mac-os-x-utilities-with-gnu-core-utilities). Maybe that's what they do on Travis. 

> Maybe they have the GNU tools by default? 

Any idea how we could reverse this?

John-Colvin:
Made a mistake above, `find` is actually in `findutils`. `findutils` does have `--with-default-names`.

We can do `brew uninstall --force findutils <etc>` before testing.

wilzbach:
cool, let's see what happens :O
https://github.com/dlang/installer/pull/337

PR #336, opened 2018-07-31
Use language:generic for Travis builds

wilzbach:
According to https://github.com/travis-ci/travis-ci/issues/4895#issuecomment-150703192o

To avoid issues like https://github.com/dlang/installer/pull/335 or https://github.com/dlang/installer/pull/334

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: c
+language: minimal
 
 addons:
   apt:


ZombineDev:
`generic` still comes with a lot of pre-installed software. If anything you may be interested in minimal, see: https://github.com/travis-ci/docs-travis-ci-com/issues/910#issuecomment-356915625

ZombineDev:
Actually reading through https://docs.travis-ci.com/user/reference/osx/ I'm not sure if specifying a different language would be of any help.

PR #337, opened 2018-07-31
OSX: uninstall brew findutils

wilzbach:
https://github.com/dlang/installer/pull/335#issuecomment-409307499


wilzbach:
CC @John-Colvin. Have a look at https://travis-ci.org/dlang/installer/jobs/410453240

GNU find supports `--version`, so I assume that GNU find was never installed?
`brew` also doesn't say anything like "uninstall successful".

John-Colvin:
I wonder what `which find` gives. I'd also consider a sanity check of running a simple `find -printf` command directly in `.travis.yml`

wilzbach:
Closing as this never worked.

PR #338, opened 2018-08-13
Install the keyring if it's missing

wilzbach:
In case the user deleted it manually, the installer should detect this.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -345,7 +345,7 @@ run_command() {
     case $1 in
         install)
             check_tools curl
-            if [ ! -f "$ROOT/install.sh" ]; then
+            if [ ! -f "$ROOT/install.sh" ] || [ ! -f "$ROOT/d-keyring.gpg" ] ; then
                 install_dlang_installer
             fi
             if [ -z "${2:-}" ]; then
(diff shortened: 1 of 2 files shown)


marler8997:
This fixed an issue I was having (https://issues.dlang.org/show_bug.cgi?id=19100). I downloaded `install.sh` directly to `~/dlang/install.sh`, so this "conditional" would always evaluate to `false` meaning it would never download `d-keyring.gpg`.  This would cause an "invalid signature" error.

marler8997:
Note: if the keyring ever changes (i.e. if someone has an old keyring installed), the install.sh script should probably detect this and remove/download the new keyring.  You could do this by checking the sha of the keyring to detect if it is out-of-date.

wilzbach:
Added a small test to it and merging now as this is a very trivial fix.

wilzbach:
> Note: if the keyring ever changes (i.e. if someone has an old keyring installed), the install.sh script should probably detect this and remove/download the new keyring. You could do this by checking the sha of the keyring to detect if it is out-of-date.

Yep, very valid point (in fact there should also be a more sophisticated process for the `install.sh` script to update itself).
I created an issue, s.t. it's not forgotten: https://issues.dlang.org/show_bug.cgi?id=19166

marler8997:
Lol...already created one :) https://issues.dlang.org/show_bug.cgi?id=19165

I marked yours as duplicate

MartinNowak:
Doesn't `~/dlang/install.sh update` also replace the keyring?
Auto-updating would be nice, maybe S3 provides ETags that could be queried?
Or we publish a timestamp under dlang.org/install.sh.mtime. Just some ideas.

PR #341, opened 2018-08-17
add changelog for signed windows binaries

MartinNowak:


MartinNowak:
No doc test here apparently.

PR #346, opened 2018-10-21
Switch from MinGW to MinGW-w64

kinke:
* Proper support for Win64
* Way more libs (900 for x64, 591 for x86 - currently around 115)
* No stdcall mangling in 64-bit .def files
* ~~Partial (?) support for UCRT~~ Support for Visual C++ 2002-2015

[I'm trying to get these libs working with LDC too, which will require msvcr120 (VS 2013).] Pinging @rainers.


kinke:
You (DMD) should probably switch to `msvcr120.lib` too. IIRC, druntime assumes that as min version in various places, leading to linking errors (e.g., math symbols) or wrong results otherwise, as MS added more-or-less-complete C99 compatibility with VS 2013, and some more additions/fixes with VS 2015.

kinke:
Currently missing symbols for a std.stdio-hello-world with LDC:
```
msvcr120.lib(msvcrt_atexit.obj) : error LNK2005: atexit already defined in msvcr120.lib(MSVCR120.dll)
phobos2-ldc.lib(gzlib.c.obj) : error LNK2019: unresolved external symbol __stdio_common_vsprintf referenced in function snprintf
phobos2-ldc.lib(file.obj) : error LNK2019: unresolved external symbol __chkstk referenced in function _D3std4file6getcwdFNeZAya
phobos2-ldc.lib(parallelism.obj) : error LNK2001: unresolved external symbol __chkstk
druntime-ldc.lib(dmain2.obj) : error LNK2001: unresolved external symbol __chkstk
phobos2-ldc.lib(crc32.c.obj) : error LNK2001: unresolved external symbol __GSHandlerCheck
phobos2-ldc.lib(inftrees.c.obj) : error LNK2001: unresolved external symbol __GSHandlerCheck
phobos2-ldc.lib(trees.c.obj) : error LNK2001: unresolved external symbol __GSHandlerCheck
phobos2-ldc.lib(crc32.c.obj) : error LNK2019: unresolved external symbol __security_check_cookie referenced in function crc32_combine_
phobos2-ldc.lib(inftrees.c.obj) : error LNK2001: unresolved external symbol __security_check_cookie
phobos2-ldc.lib(trees.c.obj) : error LNK2001: unresolved external symbol __security_check_cookie
phobos2-ldc.lib(crc32.c.obj) : error LNK2019: unresolved external symbol __security_cookie referenced in function crc32_combine_
phobos2-ldc.lib(inftrees.c.obj) : error LNK2001: unresolved external symbol __security_cookie
phobos2-ldc.lib(trees.c.obj) : error LNK2001: unresolved external symbol __security_cookie
phobos2-ldc.lib(inftrees.c.obj) : error LNK2019: unresolved external symbol __report_rangecheckfailure referenced in function inflate_table
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol set_terminate referenced in function _d_throw_exception
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol __processing_throw referenced in function _D3ldc7eh_msvc17msvc_eh_terminateFNbZv
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol __current_exception referenced in function _d_eh_swapContext
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol __current_exception_context referenced in function _d_eh_swapContext
hello.exe : fatal error LNK1120: 10 unresolved externals
```

* `at_exit()` is defined in one of the extra object files merged into the msvcrt import libs. Maybe the old MinGW .def was incomplete?
* `__stdio_common_vsprintf`: Phobos' zlib was compiled with VS 2017. MS uses inline stdio functions starting with VS 2015 (for printf etc.); fixing this will require some tricks to make it work with both older and newer MS libs.
* `__chkstk`: emitted by LLVM when coming across an `alloca`, to detect stack overflows.
* `__security_cookie`, `__report_rangecheckfailure`, `__GSHandlerCheck`: I guess something similar, but emitted by the MS C compiler; can presumably be omitted via compiler switch.
* EH will need some work; the LDC implementation currently requires VS 2015+. Rainer knows all about it, as he implemented it. ;)

rainers:
Looks pretty good to me, too.

> * `atexit()` is defined in one of the extra object files merged into the msvcrt import libs. Maybe the old MinGW .def was incomplete?

AFAICT `atexit` must exist in each DLL (if not for explicit calls, at least to allow destruction of C++ globals), Not sure why it's also exported by the MS-DLLs.

> 
> * `__stdio_common_vsprintf`: Phobos' zlib was compiled with VS 2017. MS uses inline stdio functions starting with VS 2015 (for printf etc.); fixing this will require some tricks to make it work with both older and newer MS libs.

I guess it used to be compiled with VS2013. legacy_stdio_defiitions.lib probably doesn't help here, as it is for the reverse case.

> 
> * `__chkstk`: emitted by LLVM when coming across an `alloca`, to detect stack overflows.
> 
> * `__security_cookie`, `__report_rangecheckfailure`, `__GSHandlerCheck`: I guess something similar (buffer overflow), but emitted by the MS C compiler; can presumably be omitted via compiler switch.

/GS- ? If supplied should probably included in the library similar to atexit.

> * EH will need some work; the LDC implementation currently requires VS 2015+. Rainer knows all about it, as he implemented it. ;)

Could probably be also made to work with VS2013, seems to be part of msvcr120.dll.

kinke:
With the 2nd commit and some adaptations (incl. `/GS-` for zlib compilation and linking in `ntdll.lib` for `__chkstk`), I'm now down to:
```
phobos2-ldc.lib(gzlib.c.obj) : error LNK2019: unresolved external symbol __stdio_common_vsprintf referenced in function snprintf
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol set_terminate referenced in function _d_throw_exception
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol __processing_throw referenced in function _D3ldc7eh_msvc17msvc_eh_terminateFNbZv
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol __current_exception referenced in function _d_eh_swapContext
druntime-ldc.lib(eh_msvc.obj) : error LNK2019: unresolved external symbol __current_exception_context referenced in function _d_eh_swapContext
```

I tried mocking out these few symbols via `extern (C) __gshared char` globals in `hello.d`; linking succeeds, but when trying to run it, a Windows msgbox pops up:

```
The procedure entry point localtime could not be located in the dynamic link library C:\LDC\ninja-ldc\hello.exe
```

I guess the generated .lib files aren't fully okay; e.g., there's `localtime=_localtime64` in `msvcr120.def`.

kinke:
> I guess we need to compile a C object for these defs/libs.

lib.exe and link.exe don't support undocumented `/alternatename:alias=symbol` in the command-line directly. Hacking it into an object file requires dragging it in and is so no option for each lib.

What we can definitely do is extract these special MinGW directives and auto-generate `oldnames.{c,lib}` from it, which should cover way more symbols then. E.g., `open=_open` is currently missing; current manual `oldnames.c` defines an apparently wrong alias `_open=__open`.

kinke:
I'm now more interested in getting VS 2015+ working. I looked up the VS 2015 library changes again; it introduced hexadecimal floating-point parsing, fixed accuracy of fp parsing, changed to C99 `nan` and `inf` strings, fixed vulnerability of (_)vsnprintf etc. etc.
It also wouldn't require touching LDC's EH code in druntime and neither handling the stdio thingy (I'd have an idea for that but would rather not go down that path due to the numerous VS 2015 improvements).
There are some issues:

* There's only `vcruntime140_app.def`; it seems to almost perfectly correspond to a .def auto-generated from the VS 2015 .dll though (well, after fixing the `LIBRARY` directive, I haven't found any `vcruntime140_app.dll`). A .def can be generated from any DLL by using the gendef tool included in MinGW-w64 (src only, needs to be built in MSYS; `configure` + `make`; I can upload a prebuilt binary if there's interest).
=> Easy.
* The UCRT with its large number of tiny app-ms-win-* DLLs is kinda strange. The generated .defs seem to indicate that all (?) exported functions in these tiny libs actually just forward to ucrtbase.dll. Using the MinGW-w64 ucrtbase.def doesn't really work though (e.g., no `cos` export, apparently because marked as `DATA`). The auto-generated .def (from latest Win10 SDK) differs in quite a lot of places, so I gave that one a shot, looking much better. So in essence, I replaced `msvcr120.lib` by `ucrtbase.lib` + `vcruntime140.lib` + a manual `legacy_stdio_definitions.obj`. The binary will have a direct dependency on ucrtbase.dll instead of the various app-ms-win-* DLLs.
=> Will probably require adding auto-generated 32/64-bit ucrtbase.def files; could probably be upstreamed to MinGW-w64 then too.
* When merging the current `msvcrt_stub.obj` etc. into some lib (I used vcruntime140.lib), the only remaining unresolved symbol is `__getmainargs`. And that seems like the hardest part, replicating the startup/shutdown code, which may have changed considerably with UCRT. ucrtbase seems to export a `_configure_{wide,narrow}_argv` function, but I guess the CRT sources will need to be studied to come up with a proper minimal version. Any help would be greatly appreciated. ;)

kinke:
For testing in naked `cmd.exe`:
```
$ set VSINSTALLDIR=1 // make LDC skip VS auto-detection
$ ldc2 hello.d -L-LC:\LDC\installer\windows\mingw\lib64 -mscrtlib=ucrtbase -linker="C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.15.26726\bin\Hostx64\x64\link.exe" -L/NODEFAULTLIB:vcruntime -Lvcruntime140.lib -Lntdll.lib
ucrtbase.lib(msvcrt_stub1.obj) : error LNK2019: unresolved external symbol __getmainargs referenced in function mainCRTStartup
hello.exe : fatal error LNK1120: 1 unresolved externals
```

kinke:
Alright, somewhat closer - no linking errors anymore, C cmdline parsing works too. I only later realized that druntime's _d_run_main() parses the cmdline itself so that it's probably in vain. Anyway, current state is that the executable crashes at startup, at the first GC collection (right before running the module ctors).

The dirty hacked-up `msvcrt_stub.c` getting me there:
```diff
@@ -1,4 +1,6 @@
 #include <windows.h>
+#include <fcntl.h>
+#include <stdlib.h>

 #define _CRTALLOC(x) __declspec(allocate(x))

@@ -61,8 +63,9 @@ DllMainCRTStartup (HANDLE hDll, DWORD dwReason, LPVOID lpReserved)

 #else // _APPTYPE != __UNKNOWN_APP

-extern int    _argc;
-extern char **_argv;
+extern int    __argc;
+extern char **__argv;
+extern wchar_t **__wargv;

 /* In MSVCRT.DLL, Microsoft's initialization hook is called __getmainargs(),
  * and it expects a further structure argument, (which we don't use, but pass
@@ -71,6 +74,19 @@ extern char **_argv;
 typedef struct _startupinfo { int mode; } _startupinfo;
 extern void __getmainargs( int *argc, char ***argv, char ***penv, int glob, _startupinfo *info );

+extern int _initialize_narrow_environment();
+extern char **_get_initial_narrow_environment();
+extern wchar_t **_get_initial_wide_environment();
+extern char *_get_narrow_winmain_command_line();
+extern int _configure_narrow_argv(int);
+
+typedef enum _crt_argv_mode
+{
+    _crt_argv_no_arguments,
+    _crt_argv_unexpanded_arguments,
+    _crt_argv_expanded_arguments,
+} _crt_argv_mode;
+
 /* The function mainCRTStartup() is the entry point for all
  * console/desktop programs.
  */
@@ -87,27 +103,32 @@ void WinMainCRTStartup(void)
     /* The MSVCRT.DLL start-up hook requires this invocation
      * protocol...
      */
+/*
     char **envp = NULL;
     _startupinfo start_info = { 0 };
     __getmainargs(&_argc, &_argv, &envp, 0, &start_info);
+*/

     _initterm(__xi_a, __xi_z);
     _initterm(__xc_a, __xc_z);

 #if _APPTYPE == __CONSOLE_APP
-    nRet = main(_argc, _argv, envp);
+    _set_fmode(_O_TEXT);
+    _configure_narrow_argv(_crt_argv_unexpanded_arguments);
+    _initialize_narrow_environment();
+    nRet = main(__argc, __argv, _get_initial_narrow_environment());
 #else
     {
         STARTUPINFOA startupInfo;
         GetStartupInfoA(&startupInfo);
         int showWindowMode = startupInfo.dwFlags & STARTF_USESHOWWINDOW
                            ? startupInfo.wShowWindow : SW_SHOWDEFAULT;
-        PCSTR lpszCommandLine = GetCommandLineA();
+        PCSTR lpszCommandLine = _get_narrow_winmain_command_line();
         nRet = WinMain((HINSTANCE)&__ImageBase, NULL, lpszCommandLine, showWindowMode);
     }
 #endif

-    term_atexit();
+    //term_atexit();
     _initterm(__xp_a, __xp_z);
     _initterm(__xt_a, __xt_z);
```

kinke:
Finally got it to work (Win64) after patching `msvcrt_data.c` (different segment for TLS data). I still have to clean up my local mess before pushing some commits.

kinke:
I found the actual reason for the seemingly incomplete `ucrtbase.def.in` file - they hide some functions which they override in their MinGW runtime, mainly math functions, by appending a `DATA` suffix. I revert that, so there's no more need to override full .def files.

kinke:
```
$ ldc2 -L-LC:\LDC\installer\windows\mingw\lib64 -mscrtlib=vcruntime140 -linker="C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.15.26726\bin\Hostx64\x64\link.exe" -L/NODEFAULTLIB:vcruntime -Lntdll.lib -run ..\hello.d
Hello LDC2 8
```

kinke:
Dealing with the stdcall/fastcall mangling mess is necessary for 32-bit. Hello-word links and starts successfully with LDC (still using `vcruntime140`), but segfaults a bit later on in memcpy; that is possibly related to the `_chkstk` implementation in LLVM's compiler-rt lib though, as it happens when accessing memory allocated via `alloca` [and I had troubles with the 64-bit variant too yesterday; that is finally working now].

kinke:
The 32-bit EH issue with LLD [seems related to SafeSEH](http://lists.llvm.org/pipermail/llvm-dev/2018-August/125427.html). I'm using `/SAFESEH:NO` for Win32 with both linkers, as there are 2 ASM files without them (`msvcrt_abs.asm` here and LLVM's `__chkstk` implementation), and the MS linker needs that flag for successful linking.

kinke:
Alright, got this to work too now, see last 2 commits. I additionally had to hack the LLVM _chkstk asm file, to make that object file marked safe-EH compatible (as clang has no `/safeseh` option and no equivalent AFAICT, and `ml` cannot be used due AT&T syntax...), by injecting this directly under `.text`:
```
.def    @feat.00;
.scl    3;
.type   0;
.endef
.globl  @feat.00
.set    @feat.00, 1
```

kinke:
With https://github.com/ldc-developers/ldc/pull/2886, these libs now apparently work fine with just `ldc2 {-link-internally,-linker=...\link.exe} -L-LC:\LDC\installer\windows\mingw\lib{32,64} -mscrtlib=vcruntime140` in a naked `cmd.exe` with env variable `VSINSTALLDIR=1`.

=> This should be complete from my side. I'll adapt LDC's MSVC-detection logic and then bundle these libs with LDC v1.13.

ZombineDev:
Ping @rainers @MartinNowak @WalterBright can you guys take a look?

rainers:
Good stuff! Looks pretty good to me, some comments:
- druntime has only 165 files in core.sys.windows, I suspect there are no declarations for all the additional libraries. AFAICT the D files base on the old mingw sources, too.
- to be a drop-in replacement for existing dmd distributions, msvcr100.lib should be renamed (or copied) to msvcrt100.lib.
- I noticed quite a few symbols in msvcr100.dll and msvcrt100.dll, but not in the msvcr100.lib generated by this PR, e.g. `wctob`. The latter has more symbols, though, but mostly C++.

kinke:
`wctob` is marked as `DATA` in the .def - but only in msvcr90 and msvcr100; it's exported properly in 110, 120 and UCRT.

What's the reason for holding on to VS 2010? Is that runtime preinstalled on Windows 7+ or something like that?

rainers:
>  Ah, they have their own `wctob` implementation in their `mingw-w64-crt` library. Patching out their math overrides (80-bit `long double`...) was rather simple (common macro pattern in .def.in files), but doing so for _all_ of their overrides is of course not feasible.

wctob might not be the most popular function, but it seems to affect quite a few others aswell, e.g. sqrt. I guess some will miss these.

> What's the reason for holding on to VS 2010? Is that runtime preinstalled on Windows 7+ or something like that?

IIRC it was the latest version the old mingw-project provided. It is also more likely to be already installed on older systems, but I don't think any of these are actually preinstalled (but msvcrt.dll, i.e. VC7). The druntime/phobos makefiles also default to building against VS 2010, but I guess nobody but Walter uses that default.

I agree it is a plus to provide the platform import libraries, but it feels incomplete without the respective definitions in druntime. Being able to choose the VS runtime is also great, but if you are getting that serious with diving into D, you will likely want to install the static or debug libraries aswell.

I don't mind a few extra MB for the installer and I'd be fine with switching to VC12 as the default in the long run, but having the names match would make the transition easier and doesn't need adapting both dmd and the installer right now.

I'm not too happy about the chosen name `msvcrt100.lib`, too, but there is no precedent for `msvcr100.lib` aswell:
- Microsoft calls the library `msvcrt.lib`
- It is not a simple import library of the respective DLL, but adds some code, too.

kinke:
> wctob might not be the most popular function, but it seems to affect quite a few others aswell, e.g. sqrt. I guess some will miss these.

Yeah, but figuring out these missing ones and adding appropriate fixups in `sanitizeDef()` will be very tedious, especially if LDC is using VC14 and DMD VC10.
I've just tried linking the Phobos and druntime test runners with the VC14 libs with LDC; Phobos works fine, druntime complained about missing `vfw32.lib`. That lib is apparently a merge of 3 other libs; there's a corresponding .mri file in the MinGW-w64 src tree. Luckily, that one seems to be the only relevant special case, so I'll add that merge as additional step.

> I agree it is a plus to provide the platform import libraries, but it feels incomplete without the respective definitions in druntime.

Well I don't think the DirectX bindings for example should be in druntime. But it's IMO nice that you can use 3rd party ones (https://forum.dlang.org/thread/pb170v$1bkq$1@digitalmars.com) and still link without having to install the MS toolchain.

> Being able to choose the VS runtime is also great, but if you are getting that serious with diving into D, you will likely want to install the static or debug libraries aswell.

As for debug libs, there are `msvcr90d` and `msvcr120d` too, but only those 2. ;)
Sure, serious development will still require the MS libs, as those are guaranteed to be complete and correct and offer the static-linking option.
The different msvcr* versions don't add a lot to the package size, they're around 300 KB each.

> having the names match

Sure, no problem from my side if that's easier for DMD; LDC uses `vcruntime140` anyway. [The names without trailing 't' would match the DLL names (8.3 format...).]

kinke:
I felt motivated enough to fixup almost all missing functions in `ucrtbase.lib` and `msvcr100.lib`.

kinke:
> I've just tried linking & running the Phobos and druntime test runners (x64 only) with the VC14 libs with LDC

Retested with new libs, Win64 and Win32, all tests pass.

kinke:
> The debug runtimes are not distributable, so it doesn't make much sense to add these, does it?

Well if you really care about those non-compressed ~1.4 MB (32/64 bit combined), it would probably make sense to remove all runtimes < VC10 too (plus `msvcr120_app`, whatever that is - Windows store apps?), that would at least save about 3.2 MB combined.

kinke:
> I still see a couple of minor differences on not so common symbols

I used the gendef tool to generate the reference .def, as noted in the commit msg, and added all symbols which were marked with `DATA`. There shouldn't be more than half a dozen remaining missing functions, and those didn't look important at all.

> and some of the symbols we declared so far are not even exported by the DLL

That's still true for these libs too, e.g., some SSE2 functions are supposedly exported by the 64-bit lib, while they are actually only available in the 32-bit one, at least according to the reference .ref.

The old libs also nastily defined some invalid aliases for Win64, like `time=_time32` IIRC, and stuff like that. So these libs should be a pretty big improvement for Win64 (which is much more widely used than Win32 for LDC according to download numbers).

These libs will be bundled with the imminent v1.13-beta1 release, so feel free to wait for some feedback before using them with DMD too.

kinke:
Btw, your bundled LLD should probably be upgraded to v7. That version is the first claiming production-readiness for Windows targets.

kinke:
Heh, just noticed that the old libs don't include `vfw32.lib`. So apparently nobody tested linking the druntime test runner with those libs. ;)

rainers:
> Heh, just noticed that the old libs don't include `vfw32.lib`. So apparently nobody tested linking the druntime test runner with those libs. ;)

Not sure what the test_runner is used for but to slow down the tests and pollute the output. It's not run on Windows, but it also doesn't need any special library AFAICT.

rainers:
> Well if you really care about those non-compressed ~1.4 MB (32/64 bit combined), it would probably make sense to remove all runtimes < VC10 too (plus `msvcr120_app`, whatever that is - Windows store apps?), that would at least save about 3.2 MB combined (still less than 10% of the overall size though).

It's not about the size, but the maintenance and documentation burdon and possible support requests what to do with these.

rainers:
> Well, there's https://github.com/dlang/druntime/blob/master/src/core/sys/windows/vfw.d#L14, and at least LLD complains if it cannot find a lib.

Ah, ok. Win64 probably fails without the library, but noone has tested that combination. For OMF, optlink just emits a warning for missing libraries.

kinke:
I was too optimistic about the .def files containing all aliases required to generate `oldnames.lib` (https://github.com/ldc-developers/ldc/issues/2900). Fixed now.

kinke:
> The old libs also nastily defined some wrong aliases for Win64, like time=_time32 IIRC, and stuff like that.

I should have checked thoroughly (MSDN was a bit misleading; their C headers default to `_localtime64`); the old aliases were correct, the MinGW ones are wrong => https://github.com/ldc-developers/ldc/issues/2903.

After seeing that there were more wrong aliases like this (sometimes even in the wrong order - `realName == weakName`) and a whole lot of missing ones, I now ignore the MinGW aliases altogether, wrote a little tool (`extractAliases.d`) to extract the aliases from the MS libs, and use files based on the VS 2015 libs as input for the generated `oldnames.lib`, see the last commit.

This is taking much more time and persistence than anticipated. ;)

wilzbach:
So what's missing for this to be integrated with DMD?

kinke:
Nothing from my point of view, but as I said, you may want to wait a bit for some more feedback. People reported two issues very soon after releasing v1.13 beta1, those have been fixed; no more reports after beta2, but I assume a fair share of users will wait for the final.

kinke:
Well, someone should probably run the DMD tests with those libs before shipping with them; LDC uses the 2015 runtime version, while you'll apparently want to stick with the 2010 one (which I don't understand, but I also don't understand why COFF isn't the default for Win32 targets, or why DMD doesn't ship with 64-bit Windows executables - it's 2019 soon).

Ah and Rainer was afraid of support liability when providing other runtime versions too; I guess that's not a real issue, it's all undocumented and 95+% of Windows users probably wouldn't know how to select a different one.

wilzbach:
We're now using Azure to build these libraries and have thus unified all builds back to master.
Hence, I went ahead and retargeted master and rebased for you.

wilzbach:
It's now up at http://downloads.dlang.org/other (-> http://downloads.dlang.org/other/mingw-libs-6.0.0.zip)

wilzbach:
> We still need to integrate it with DMD

Integration would be: https://github.com/dlang/installer/pull/377

kinke:
Thanks!

PR #350, opened 2018-10-26
update install.sh script for new ddo nightlies

MartinNowak:
- replace all nightlies.dlang.org URLs with downloads.dlang.org/nightlies
- quite some changes due to losing an alternative HTTPS download source
- rewrite dmd-2018-10-14 to dmd-master-2018-10-14 in the script so to
  not clutter the download page with plenty of confusing symlinks

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -406,11 +406,11 @@ install_dlang_installer() {
     tmp=$(mkdtemp)
     local mirrors=(
         "https://dlang.org/install.sh"
-        "https://nightlies.dlang.org/install.sh"
+        "https://s3-us-west-2.amazonaws.com/downloads.dlang.org/other/install.sh"
     )
     local keyring_mirrors=(
         "https://dlang.org/d-keyring.gpg"
-        "https://nightlies.dlang.org/d-keyring.gpg"
+        "https://s3-us-west-2.amazonaws.com/downloads.dlang.org/other/d-keyring.gpg"
     )
 
     mkdir -p "$ROOT"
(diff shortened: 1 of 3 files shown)


script/install.sh
```diff
@@ -406,19 +406,28 @@ install_dlang_installer() {
...
+        # use fixed commit hash to protect against tampering of the repo when we
+        # don't yet have an old keyring to verify the new one
+        "https://github.com/dlang/dlang.org/raw/7dfae2f8a4c33a2471767cd5c2e45a5e7da4ef10/d-keyring.gpg"
+    )
+    local keyring_mirrors_verify=(
+        "https://dlang.org/d-keyring.gpg"
+        # use possibly insecure but up-to-date mirrors when we have an old
+        # keyring to verify the download
+        "http://downloads.dlang.org/other/d-keyring.gpg"
+        "https://github.com/dlang/dlang.org/raw/master/d-keyring.gpg"
```
wilzbach: We haven't uploaded a signature of the current keyring from a previous one into GitHub, so this would always fail, no?

travis.sh
```diff
@@ -45,6 +42,13 @@ if [ "${TRAVIS_OS_NAME:-}" != "osx" ]; then
     )
 fi
 
+if [ "${TRAVIS_OS_NAME:-}" != "osx" ]; then
+    # check that fixed d-keyring.gpg commit hash is up-to-date
+    KEYRING_SHA=$(curl -fsSL 'https://api.github.com/repos/dlang/dlang.org/commits?sha=master&path=d-keyring.gpg' | jq '.[0].sha' --raw-output)
+    grep -F "$KEYRING_SHA" ./script/install.sh
```
wilzbach: Nice :+1:

wilzbach:
> quite some changes due to losing an alternative HTTPS download source

How about including the initial keyring with the download script? (this would also make updating the script + keyring easier as then only the script needs to be updated)

script/install.sh
MartinNowak: Here is the signature https://github.com/dlang/dlang.org/commit/7dfae2f8a4c33a2471767cd5c2e45a5e7da4ef10#diff-bcd49c888a108fa6e5ad8b14270bafd5.

MartinNowak:
Found a simpler solution using https://s3-us-west-2.amazonaws.com/downloads.dlang.org/other/d-keyring.gpg.

wilzbach:
Isn't Route 53 very expensive?

ZombineDev:
https://aws.amazon.com/route53/pricing/

Is it?

MartinNowak:
@ZombineDev Perfect Is The Enemy Of Good

We only have so much control over the domain dlang.org (administrated by Jan) our S3 bucket (owned by Brad). Given the download sizes and that it's only for installation, geo-proximity seems less of an issue.

This endpoint would have never changed if I had access to downloads.dlang.org back when installing the nightly service, serving them from the build server was just a workaround. We haven't changed any download URL in years.

Given that things work quite OK atm., I don't think it's worthwhile yet (for us) to touch downloads.dlang.org.
If someone wants to work on a better CDN setup for downloads.dlang.org I'd highly welcome that effort though. There are a couple of issues, most prominently the lack of HTTPS support.

MartinNowak:
You could have just deployed the script @wilzbach.

ZombineDev:
I'm asking because the downloads are frustratingly slow on my 200Mbps connection. This was the by far the slowest part of a docker recipe which downloaded and built OpenSSL and vibe.d from source, among other things :O

PR #351, opened 2018-11-03
Build lld from LLVM 7.0.0

rainers:


wilzbach:
Currently with bad internet, but I will to to update it in the S3 bucket on Monday.

kinke:
The 7z archive of v2.084.0-beta.1 still ships with LLD v5.0.1. I bet the newer version would fix https://forum.dlang.org/thread/mailman.5899.1545250672.29801.digitalmars-d-bugs@puremagic.com; I recall issues with previous LLD versions if there was a .pdb file already, but haven't seen those errors in a while.

wilzbach:
Thanks a lot for the ping!
Apparently this needs a manual update: https://github.com/dlang/installer/pull/355

It's not visible via the HTML pages as I haven't managed to get the newest version of downloads.dlang.org to work yet, but the file and its signature are in the bucket:  http://downloads.dlang.org/other/lld-link-7.0.0.zip

wilzbach:
> It's not visible via the HTML pages as I haven't managed to get the newest version of downloads.dlang.org 

So I figured out how to use https://github.com/dlang/downloads.dlang.org and the HTML index pages are updated now:

MartinNowak:
> > 
> 
> So I figured out how to use https://github.com/dlang/downloads.dlang.org and the HTML index pages are updated now

That's part of the release process, so you could've found it here https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0#file-release-sh-L161-L162.

PR #367, opened 2019-03-03
Issue 19716 Add MANPATH to activate(.fish)

bolry:
Add the MANPATH environment variable to both the `activate` and `activate.fish` generated scripts in `install.sh`. Gives access to man pages for, e.g., dmd and dmd.conf.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -381,7 +381,7 @@ run_command() {
             else
                 log "
 Run \`source $ROOT/$2/activate${suffix:-}\` in your shell to use $2.
-This will setup PATH, LIBRARY_PATH, LD_LIBRARY_PATH, DMD, DC, and PS1.
+This will setup PATH, LIBRARY_PATH, LD_LIBRARY_PATH, MANPATH, DMD, DC, and PS1.
 Run \`deactivate\` later on to restore your environment."
             fi
             ;;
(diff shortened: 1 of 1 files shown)


wilzbach:
Thanks a lot for your PR, but could you please add a simple test for it?
Check travis.sh for details.

bolry:
Will do. Travis is new to me but I'll have a look.

wilzbach:
It's basically just one big shell script which we run in the CI. Let me know if you need help.

bolry:
Seemingly I need some ideas about the travis runtime environment to check it. Maybe `man` isn't available?

wilzbach:
I presume LDC just doesn't come with a man page or has it somewhere else?

travis.sh
```diff
@@ -75,6 +75,16 @@ do
     command -v dub >/dev/null 2>&1 || { echo >&2 "DUB hasn't been installed."; exit 1; }
     deactivate
 
+    # Check man pages do not destroy system man pages availability
+    if man ls >/dev/null 2>&1; then man_ls_exit_status=0; else man_ls_exit_status=$?; fi
```
wilzbach: This is a bit of an overkill. `set -e` is used in the script, so you don't need the if else at all. The script will fail of the command fails and you can use `! ..` for negation.

bolry: Point of the matter is that `man dmd` before setting up MANPATH thru `activate` should cause the exit code to be `16` and I didn't know any better way to check that. So, the two first checks are just to pick up the current exit codes before `activate` to compare with after `deactivate`. The middle two checks are 1) `dmd`'s man page available, 2) we haven't lost the system man pages (thus unchanged from the first test).

Another thing, I didn't see any tests in travis for using the fish-shell. Should I introduce `man` page testing for that shell too (is it available?).

BTW, LDC means ...?

bolry:
I've changed to only add `MANPATH` for dmd compilers.

AndrewEdwards:
@bolry Request review and resolution of any conflicts associated with this pull request. Let's whip this in shape for merger.

AndrewEdwards:
@wilzbach, I doubt @bolry is with us anymore. Do you mind taking over and seeing this to completion?

PR #370, opened 2019-04-13
Build LLD with Azure on master

wilzbach:
Currently there are a few branches at this repository:

- [MinGW libs](https://github.com/dlang/installer/tree/build-mingw-libs)
- [LLD](https://github.com/dlang/installer/tree/build-lld) - the LLVM linker
- [Curl](https://github.com/dlang/installer/tree/build-curl)

The reason was to save resources as AppVeyor only provides one parallel build. Imho, now that we switched DMD to Azure, we can also switch to Azure here and unify everything in master.


andralex:
Thanks! I have no idea what I'm doing but I assume this is the right thing!

wilzbach:
This was _not_ the right thing, sorry. We needed to activate Azure first and test whether it actually passes, but I like your enthusiasm and we can fix it up later, no harm done ;-)
Thanks!

PR #372, opened 2019-04-13
Build curl on Azure

wilzbach:


MartinNowak:
Because of the fancier integration with github?

wilzbach:
Because
1) they have ten parallel jobs for free
2) they don't require you to to login for every project with a different account
3) this is the last use of Appveyor in the dlang ecosystem (DMD already switched + the other two build jobs). So it would be one different CI provider less to worry about.
...

wilzbach:
Oh and the main motivation was to have all this code united back in master, because we no longer are so short on build resources.

MartinNowak:
> Oh and the main motivation was to have all this code united back in master, because we no longer are so short on build resources.

We only update the other projects every few months though and building lld in particular takes quite a while, so leaving this in separate branches instead of wasting hours of compute (and lots of unneeded artifacts) still makes sense to me.

PR #377, opened 2019-04-14
Upgrade MinGW to 6.0.0

wilzbach:
Upgrade the installer to use the Mingw 6 definitions from https://github.com/dlang/installer/pull/346.
Would be great if someone can give this a quick test, s.t. we can ship with with the upcoming 2.086 beta.

This could fix:
- https://issues.dlang.org/show_bug.cgi?id=19773

### DIFF
--- a/changelog/mingw-6.dd
+++ b/changelog/mingw-6.dd
@@ -1,0 +1,3 @@
+The bundled Windows libraries have been changed from MinGW to MinGW-w64.
+
+The bundled Windows libraries and definitions have been changed from MinGW 5.0.2 to MinGW 7.0.0.
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -491,7 +491,7 @@ int main(string[] args)
     enum libC = "snn.lib";
     enum libCurl = "libcurl-7.64.0-WinSSL-zlib-x86-x64.zip";
     enum omflibs = "omflibs-winsdk-10.0.16299.15.zip";
-    enum mingwlibs = "mingw-libs-5.0.2-1.zip";
+    enum mingwlibs = "mingw-libs-7.0.0.zip";
     enum lld = "lld-link-8.0.0.zip";
 
     auto oldCompilers = platforms


kinke:
Wrt. the changelog, the important thing is the switch from MinGW to MinGW-w64. They are separate projects - from their website:

> Mingw-w64 is an advancement of the original mingw.org project, created to support the GCC compiler on Windows systems. It has forked it in 2007 in order to provide support for 64 bits and new APIs. It has since then gained widespread use and distribution.

wilzbach:
Updated the description.

kinke:
As @Basile-z has reported in the [forum](https://forum.dlang.org/post/vslwlgwjykdtocmwivsf@forum.dlang.org), there are issues wrt. duplicate 'weak aliases' in druntime and these new libs. Additionally, the aliases in the MinGW-based libs are based on the VS2015 libs (for LDC), so they are likely not to fully correspond to the 2010 libs used by DMD.

So as I mentioned in the forum, there are 2 ways of tackling this:
1) Extract and use the aliases from the 2010 libs.
2) Make DMD switch to the 2015 libs by default and adapt the druntime aliases accordingly.

As there wasn't much feedback in the forum (well, to be expected I guess), I'd feel more comfortable if someone like @rainers would do some proper testing. For LDC, I tested the libs by linking and running the druntime and Phobos test runners (both 32 and 64 bits); the LDC test runners pull in every single object file, so that one can be sure there are no conflicting aliases, no missing libs referenced by `pragma(lib, ...)` and no missing symbols referenced by the unittests.

wilzbach:
Bumped to 6.0.0-1 (from https://github.com/dlang/installer/pull/388).

> As there wasn't much feedback in the forum (well, to be expected I guess), I'd feel more comfortable if someone like @rainers would do some proper testing

Yeah that would be awesome (:

rainers:
Sorry, somehow never got back to this. Just tried a simple hello-world:
```
c:\tmp>dmd -mscrtlib=msvcrt100 -m32mscoff args.d
lld-link: error: /alternatename: conflicts: _fileno=__fileno
```

kinke:
Yes, see my last comment above. There are conflicting aliases in https://github.com/dlang/druntime/blob/master/src/rt/msvc.c [and msvc_math.c] (as the ones here stem from the VS 2015 libs); LDC's druntime has adapted these little C files accordingly.

In the meantime, I'm convinced DMD should switch to the 2015/UCRT libs and give up the 2010 libs, it's 2020 by now.

rainers:
As far as I can tell, only _fileno is an issue. If I comment that symbol and build the libs with buildsdk, this compiles and links with -mscrtlib=msvcrt120 -m32mscoff:
```D
import core.stdc.stdio;
import core.stdc.math;

void test()
{
    void* p = &snprintf;
    p = &vsnprintf;
    p = &core.stdc.stdio._fputc_nolock;
    p = &core.stdc.stdio._fgetc_nolock;
    p = &rewind;
    p = &clearerr;
    p = &feof;
    p = &ferror;
    core.stdc.stdio.fileno(core.stdc.stdio.stderr);
    
    acosf(0);
    asinf(0);
    atanf(0);
    atan2f(0, 0);
    cosf(0);
    sinf(0);
    tanf(0);
    coshf(0);
    sinhf(0);
    tanhf(0);
    expf(0);
    logf(0);
    log10f(0);
    powf(0, 0);
    sqrtf(0);
    ceilf(0);
    floorf(0);
    fmodf(0, 1);
}

void main(string[] args)
{
    import std.stdio;

    writeln(args);
}
```
The generated x64 libraries have an additional `_`, though. Not sure why these are not in the zip from the release archive.

rainers:
> In the meantime, I'm convinced DMD should switch to the 2015/UCRT libs and give up the 2010 libs, it's 2020 by now.

I'm not a big fan of the newer VC runtimes when it comes to distributing your application: you either have to run the redistributable installers or need to struggle with manifests for local copies. The older runtimes can just be copied with your executable. (I think we should just add the msvcr100.dll to the dmd distribution).

It might be slightly more feasible if we don't rely on the VC libraries, but only on the UCRT as these should already exist on any recent system. Older Windows versions have been stubborn installing the UCRT, eventually (Windows 7 or earlier).

kinke:
> I'm not a big fan of the newer VC runtimes when it comes to distributing your application

I haven't gotten any complaints by LDC users in this regard yet. For ~professional redistribution without integrated redistributable installer, people would hopefully prefer the MS toolchain and static linking anyway over copying the DLLs.
And wrt. users simply wanting to compile, link and run the binaries on their own system, I guess the chance for an installed VC++ runtime 2015+ is higher than that of VS 2010 nowadays.

> It might be slightly more feasible if we don't rely on the VC libraries, but only on the UCRT

Not feasible for LDC as you know, but that might be an option for DMD.

rainers:
> The generated x64 libraries have an additional `_`, though. Not sure why these are not in the zip from the release archive.

Actually, they were wrong there, too, but the _fileno error blocked other output. See https://github.com/dlang/installer/pull/414/commits/1a76d2634b57eaedabc5c83d97998cb01bd1e0dc

rainers:
> For ~professional redistribution without integrated redistributable installer, people would hopefully prefer the MS toolchain and static linking anyway over copying the DLLs.

I hope so, too. TBH I have no use for the mingw version, other than stopping people from complaining about having to install the MS toolchain.

> And wrt. users simply wanting to compile, link and run the binaries on their own system, I guess the chance for an installed VC++ runtime 2015+ is higher than that of VC++ 2010 nowadays.

Might be true, but even more simpler if the required DLL just comes with dmd (assuming the PATH is set to include bin/bin64).

kinke:
> See 1a76d26

Ah sry, I've always run these tools in separate x64/x86 VS tools command prompts. - I should probably upstream the MinGW-w64 v7 changes from the LDC repo, so that we're based on the same MinGW definitions. They've apparently used the Wine headers and made according fixes/extensions in v7.

> TBH I have no use for the mingw version, other than stopping people from complaining about having to install the MS toolchain.

Yeah that's mostly it, but it's also nice for cross-linking Windows binaries on Posix hosts, making that substantially simpler than cross-linking to other Posix *targets* (need C cross-toolchain).

wilzbach:
Updated to 7.0.0 which should point to the artificat built in https://github.com/dlang/installer/pull/414.

Thanks a lot @rainers

PR #387, opened 2019-05-10
Installer updates: remove D1, update VisualD, rename Publisher

wilzbach:
Looked pretty straight forward.

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -39,7 +39,7 @@
 ; Routinely Update
 ; ----------------
 ; Visual D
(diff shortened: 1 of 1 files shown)


windows/d2-installer.nsi
```diff
@@ -39,22 +39,16 @@
 ; Routinely Update
 ; ----------------
 ; Visual D
-!define VersionVisualD "0.48.1"
+!define VersionVisualD "0.48.2"
```
ZombineDev:
```suggestion
!define VersionVisualD "0.49.2"
```

See: https://github.com/dlang/visuald/releases

PR #388, opened 2019-05-11
MinGW-w64 libs: Fix DLL entry point

kinke:
This should fix the issues in https://forum.dlang.org/thread/uxmbtqqdomrrsnrssapc@forum.dlang.org.

I've tested linking the following DLL (with LDC), as both x86/x64 and with both msvcrt100.lib (2010) and vcruntime140.lib (2015), and it works:
```D
import core.sys.windows.windows;

extern(Windows) BOOL DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
{
    return FALSE;
}

export void foo()
{
    import core.stdc.stdlib : malloc, free;
    auto blub = malloc(1024);
    free(blub);
}
```

### DIFF
--- a/windows/mingw/msvcrt_stub.c
+++ b/windows/mingw/msvcrt_stub.c
@@ -39,10 +39,13 @@ extern int __ref_oldnames;
 
 #if _APPTYPE == __UNKNOWN_APP
 
+extern BOOL WINAPI DllMain (HINSTANCE, DWORD, LPVOID);
+
 BOOL WINAPI
-DllMainCRTStartup (HANDLE hDll, DWORD dwReason, LPVOID lpReserved)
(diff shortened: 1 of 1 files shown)


kinke:
Do we still need something for the prebuilt package (and #377) to be updated? SHA checksums etc.?

wilzbach:
Yes, I can upload the built binary to S3 later and bump the version number. Not sure about #377 as I don't use Windows and when I ask in the Forum, some people reported issues. Were these resolved with this?

kinke:
Cheers. :) - Wrt. #377,I just meant the  version bump there too; the issue/decision there is untouched by this.

PR #399, opened 2019-08-12
fix issue 20120 - libcurl.dll hangs when running std.net.curl unittests

rainers:
disable the workaround introduced in https://github.com/curl/curl/commit/72d5e144fbc6a9db264ae425bb788af218f25d9e
as it is causing the issue and the curl developers were not really convinced

disabled by defining DONT_USE_RECV_BEFORE_SEND_WORKAROUND when compiling curl

### DIFF
--- a/windows/build_curl.bat
+++ b/windows/build_curl.bat
@@ -43,7 +43,7 @@ SET PATH=%MINGW_PATH%;%ORIG_PATH%
 call "c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
 
 mingw32-make -C zlib -f win32\Makefile.gcc || exit /B 1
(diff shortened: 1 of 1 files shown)


rainers:
Just tried the artifacts: the phobos unittests pass with the new DLL.

kinke:
Great that you found it. I always assumed it was LDC-specific (newer MSCRT etc.) as DMD happily shipped with newer versions...

wilzbach:
Deployed as http://downloads.dlang.org/other/libcurl-7.65.3-2-WinSSL-zlib-x86-x64.zip and http://downloads.dlang.org/other/libcurl-7.65.3-2-WinSSL-zlib-x86-x64.zip.sig

So for some reason the HTML index generation script is currently broken, so you can't see it in the index list yet.

wilzbach:
Update PR for 2.088: https://github.com/dlang/installer/pull/400

PR #400, opened 2019-08-20
Update curl to 7.65

wilzbach:
See also: https://github.com/dlang/installer/pull/399


wilzbach:
CC @thewilsonator and others? You still fine with shipping this in 2.088? It's a bit critical as otherwise the Phobos testsuite doesn't run properly (see e.g. https://github.com/dlang/installer/pull/399).

rainers:
Thanks @wilzbach

PR #406, opened 2019-11-28
Resolve latest version for dub as additional compiler correctly

wilzbach:


wilzbach:
Merging as this actively blocks the DUB CI.

PR #408, opened 2020-01-17
test: Don't clobber user's work in worktree

CyberShadow:
Running the test script should not destroy users' WIP changes to the install script.


CyberShadow:
Superseded by #409.

PR #409, opened 2020-01-17
script/install.sh: Add initial Windows support

CyberShadow:
Supports installing Windows binaries of DMD, LDC, and Dub on Windows, including nightlies, when running in POSIX-like environments such as MSYS.

Tests pass under MSys2 and CygWin for me.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -129,16 +129,9 @@ case $(uname -s) in
     Darwin) OS=osx;;
     Linux) OS=linux;;
     FreeBSD) OS=freebsd;;
+    *_NT-*) OS=windows;;
     *)
         fatal "Unsupported OS $(uname -s)"
-        ;;
-esac
-case $(uname -m) in
-    x86_64|amd64) ARCH=x86_64; MODEL=64;;
-    aarch64) ARCH=aarch64; MODEL=64;;
-    i*86) ARCH=x86; MODEL=32;;
-    *)
-        fatal "Unsupported Arch $(uname -m)"
         ;;
 esac
 
@@ -556,10 +549,12 @@ install_compiler() {
             fi
         fi
 
-        if [[ $ver > "2.068.0z" ]]; then
-            local arch="tar.xz"
+        if [[ $OS == windows && $ver > "2.068.0" ]] && command -v 7z &>/dev/null; then
+            local ext=".7z"
+        elif [[ $ver > "2.068.0z" && $OS != windows ]]; then
+            local ext=".tar.xz"
         else
-            local arch="zip"
+            local ext=".zip"
         fi
 
         if [[ $ARCH != x86* ]]; then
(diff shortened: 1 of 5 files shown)


script/install.sh
```diff
@@ -591,15 +586,23 @@ install_compiler() {
         if [[ $ARCH != x86* ]]; then
             fatal "no DMD binaries available for $ARCH"
         fi
-        local url="http://downloads.dlang.org/nightlies/$1/$basename.tar.xz"
+        local ext
+        test $OS = windows && ext=.7z || ext=.tar.xz
```
wilzbach: Why can't we use the existing `ext`?

CyberShadow: You mean declare `ext` once at the top of the function instead of per `if` branch?

CyberShadow:
It's green! (Travis had some troubles with its OSX VMs yesterday)

CyberShadow:
@thewilsonator

thewilsonator:
I don't have meg rights here.

CyberShadow:
That's not right. How about now?

CyberShadow:
Thanks! :)

veelo:
Wow, talk about timing! Today I started looking into D version management on Windows. First hacking on dvm, then on install.sh, so I'm very grateful for this PR and for stumbling on it :-) My initial take was using [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10) to run bash, but using MSys2 probably makes more sense. Is there an associative PR to https://dlang.org/install.html?

CyberShadow:
> First hacking on dvm

For personal use I use my tool [dver](https://github.com/CyberShadow/misc/blob/master/dver.d), but I wanted to replace AppVeyor with Travis for Digger, ergo this PR.

> My initial take was using [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10) to run bash

I think that should work. Note though that this will get you Linux D compilers which create Linux (not Windows) executables, which should still run under WSL but might not be what you want.

> Is there an associative PR to https://dlang.org/install.html?

Oh, I didn't know about that page, thanks. I will send a PR.

veelo:
I tested this using pristine MSYS2, which has neither unzip nor 7z. I tried installing dmd, following the signs, but that failed due to a problem with unzip.

The script first tries 7z, and falls back to unzip if 7z is not present. If unzip is not present either, msys prompts you to install it. So I installed unzip using `pacman -S unzip` and tried again. The dmd `.zip` gets downloaded but the process ends with the message `warning:  .installer_tmp_ALYBiK/e1bFrh/dmd.2.090.0.windows.zip appears to use backslashes as path separators`. This appears to actually be an error, not a warning, because dmd is nowhere to be found. Or maybe some other error occurred after that, but there is no sign of that, not with `-v` either.

But, installing 7z using `pacman -S p7zip` makes the script to download the dmd `.7z`, which it then successfully installs.

Using 7z will only work for versions 2.068.1 and upwards, prior to that there are only `.zip`s.

I don't know if we can make unzip to work, or whether we should special-case the script for MSYS requiring 7z and restricting version history. Does this work better under Cygwin?

CyberShadow:
On Arch Linux I get:

```
$ unzip dmd.2.090.0.windows.zip 
Archive:  dmd.2.090.0.windows.zip
warning:  dmd.2.090.0.windows.zip appears to use backslashes as path separators
  inflating: dmd2/src/phobos/std/compiler.d  
  inflating: dmd2/src/druntime/import/core/sys/linux/config.d  
  ...
```

It then continues to unpack the archive successfully.

Maybe the MSYS unzip utility is crashing or otherwise failing due to an unrelated reason?

veelo:
The Windows 10 "Extract All..." option in the file explorer has no problem extracting it either (apart from being slow).

veelo:
Hah, doing `unzip dmd.2.090.0.windows.zip` in MSYS2 manually works perfectly as well, same output as you get.

I don't know what's going on.

CyberShadow:
We invoke `unzip` with `-q`, so it would not produce a listing of each file.

How do you know it's failing?

veelo:
If I comment-out the contents of `cleanup()`, I can see that the `.zip` is properly extracted into the temporary directory. So `unzip` is not the problem. Something else is failing.

CyberShadow:
Can you post a `tree` of `~/dlang`?

Also running the script with `bash +x` will show what command is failing.

veelo:
I fixed it: #410.

PR #414, opened 2020-01-24
build mingw libraries from mingw64 7.0.0

rainers:
remove _fileno from oldnames, it is already provided by druntime


kinke:
LDC diff from master:

buildsdk.d:
```diff
@@ -91,7 +91,10 @@ void sanitizeDef(string defFile)
         // ucrtbase.def:
         "_assert", "_cabs", "_fpreset", "_tzset",
         "ceil", "ceilf", "coshf", "fabs",
+        "feclearexcept", "fegetenv", "fegetexceptflag", "fegetround", "feholdexcept",
+        "fesetenv", "fesetexceptflag", "fesetround", "fetestexcept",
         "floor", "floorf", "modf", "modff",
+        "lgamma", "lgammaf", "lgammal",
         "sinhf", "sqrt", "sqrtf", "wcsnlen",
         // additional ones in msvcr100.def:
         "__report_gsfailure",
@@ -104,7 +107,7 @@ void sanitizeDef(string defFile)
         "_wassert",
         "acosf", "asinf", "atan2", "atan2f", "atanf",
         "btowc",
-        "cos", "cosf", "exp", "expf", "fmod", "fmodf", "ldexp",
+        "cos", "cosf", "exp", "expf", "fmod", "fmodf", "frexp", "ldexp",
         "longjmp",
         "llabs", "lldiv",
         "log", "log10f", "logf",
@@ -143,10 +146,13 @@ void sanitizeDef(string defFile)
         }

         // Un-hide functions overridden by the MinGW runtime.
-        foreach (name; overriddenMinGWFunctions)
+        if (line.endsWith(" DATA"))
         {
-            if (line.length == name.length + 5 && line.startsWith(name) && line.endsWith(" DATA"))
-                return name;
+            foreach (name; overriddenMinGWFunctions)
+            {
+                if (line.length == name.length + 5 && line.startsWith(name))
+                    return name;
+            }
         }

         // Don't export function 'atexit'; we have our own in msvc_atexit.c.
@@ -248,7 +254,7 @@ bool defWithStdcallMangling2implib(string defFile)
             line.startsWith("LIBRARY ") || line.startsWith("EXPORTS"))
             return line;

-        if (line.endsWith(" DATA"))
+        if (line.length > 5 && (line[$-5] == ' ' || line[$-5] == '\t') && line.endsWith("DATA"))
         {
             fields ~= line[0 .. $-5];
             return line;
@@ -336,11 +342,11 @@ bool defWithStdcallMangling2implib(string defFile)
     return true;
 }

-void c2lib(string outDir, string cFile)
+void c2lib(string outDir, string cFile, string clFlags = null)
 {
     const obj = buildPath(outDir, baseName(cFile).setExtension(".obj"));
     const lib = setExtension(obj, ".lib");
-    cl(obj, quote(cFile));
+    cl(obj, clFlags ? clFlags ~ " " ~ quote(cFile) : quote(cFile));
     runShell(`lib "/OUT:` ~ lib ~ `" ` ~ quote(obj));
     std.file.remove(obj);
 }
@@ -431,7 +437,7 @@ void buildOldnames(string outDir)

 void buildLegacyStdioDefinitions(string outDir)
 {
-    c2lib(outDir, "legacy_stdio_definitions.c");
+    c2lib(outDir, "legacy_stdio_definitions.c", "/O2");
 }

 // create empty uuid.lib (expected by dmd, but UUIDs already in druntime)
```

legacy_stdio_definitions.c:
```diff
@@ -15,6 +15,7 @@ void __legacy_stdio_definitions()
     printf(NULL);
     scanf(NULL);
     snprintf(NULL, 0, NULL);
+    _snprintf(NULL, 0, NULL);
     sprintf(NULL, NULL);
     sscanf(NULL, NULL);
     swprintf(NULL, 0, NULL);
@@ -26,6 +27,7 @@ void __legacy_stdio_definitions()
     vprintf(NULL, NULL);
     vscanf(NULL, NULL);
     vsnprintf(NULL, 0, NULL, NULL);
+    _vsnprintf(NULL, 0, NULL, NULL);
     vsprintf(NULL, NULL, NULL);
     vsscanf(NULL, NULL, NULL);
     vswprintf(NULL, 0, NULL, NULL);
```

msvcrt_stub.c:
```diff
@@ -64,6 +64,17 @@ _DllMainCRTStartup (HINSTANCE hDll, DWORD dwReason, LPVOID lpReserved)
     return bRet;
 }

+BOOL WINAPI __DefaultDllMain (HINSTANCE hDll, DWORD dwReason, LPVOID lpReserved)
+{
+    return TRUE;
+}
+
+#ifdef _M_X64
+__pragma(comment(linker, "/alternatename:DllMain=__DefaultDllMain"));
+#else
+__pragma(comment(linker, "/alternatename:_DllMain@12=___DefaultDllMain@12"));
+#endif
+
 #else // _APPTYPE != __UNKNOWN_APP

 extern int    __argc;
```

rainers:
> LDC diff from master:

Thanks, I've added them here, too.

thewilsonator:
first commit message has a typo: buld should be build.

rainers:
> first commit message has a typo: buld should be build.

fixed.

kinke:
I've generated the {oldnames,msvcrt120}.aliases{32,64} for Visual C++ 2013 and compared them to the used ones based on VC++ 2015; looking good (a couple of missing aliases like `setjmp` and a few dozen extra aliases for exceptions and RTTI).

thewilsonator:
Is this good to go?

rainers:
> VC12 misses a couple of aliases like `setjmp` and contains a few dozen extra aliases for exceptions and RTTI

I guess that's not a problem, is it? `setjmp` isn't even declared for Windows anywhere in core.stdc.

kinke:
> I guess that's not a problem, is it?

Nope, I don't think so either ('looking good').

rainers:
> Is this good to go?

Yes, good to go. Someone has to upload the (archive inside the) artifact to downloads.dlang.org, though? @wilzbach ?

wilzbach:
> Yes, good to go. Someone has to upload the (archive inside the) artifact to downloads.dlang.org, though? @wilzbach ?

http://downloads.dlang.org/other/mingw-libs-7.0.0.zip
http://downloads.dlang.org/other/mingw-libs-7.0.0.zip.sig

rainers:
> http://downloads.dlang.org/other/mingw-libs-7.0.0.zip

Thanks, @wilzbach

rainers:
> http://downloads.dlang.org/other/mingw-libs-7.0.0.zip

@wilzbach Sorry for not noticing earlier: you renamed and uploaded the artifact directly, but azure packs the artifacts into a zip again. The zip itself contains another archive mingw-libs-7.0.0.zip now which is unexpected by the release script.

rainers:
There is still a problem with the msvcrt libs (not sure why I didn't notice earlier): if phobos is built with VS2017 (as on azure) or VS2019 (my local builds), gzlib.obj includes inline versions of snprintf and vsnprintf that forward to __stdio_common_vsprintf which is undefined. This causes simple programs using writeln to fail to link.

That might not be a problem for the official release build because an older VC compiler is used.

wilzbach:
> @wilzbach Sorry for not noticing earlier: you renamed and uploaded the artifact directly, but azure packs the artifacts into a zip again. The zip itself contains another archive mingw-libs-7.0.0.zip now which is unexpected by the release script.

Sorry about that. I recalled checking that, but I must have listed the contents of a different archive.
I updated the uploaded files (there's a `.sig` too) as no one was using it then.

kinke:
> if phobos is built with VS2017 (as on azure) or VS2019 (my local builds), gzlib.obj includes inline versions of snprintf and vsnprintf that forward to __stdio_common_vsprintf which is undefined

Those should all be in `legacy_stdio_definitions.lib` (assuming the libs here are built with VS2015+), which should be pulled in automatically.

rainers:
> Those should all be in `legacy_stdio_definitions.lib`

I think it is the other way: `__stdio_common_vsprintf` is a new symbol introduced by VC14 or later, so unlikely to be found in legacy_stdio_definitions.lib

> With what version are you testing then, Rainer? The VC12 one?

I tried msvcrt100.lib, msvcrt110.lib and msvrt120.lib. The problem is that gzlib.d is compiled with VC16 requiring the new symbol.

kinke:
> I think it is the other way

Argh yes of course. Well, just another issue with these old pre-UCRT libs, the druntime/Phobos C parts need to be compiled with an old pre-2015 Visual C compiler. And all other static C libs making use of the printf/scanf function families...

rainers:
> And all other static C libs making use of the printf/scanf function families...

If you compile C code, you'll likely have the MS libraries and don't need the mingw workaround.

kinke:
Not if you just link against some prebuilt libs distributed by someone else.

rainers:
Now I notice that building with -mscrtlib=vcruntime140 works for my example, but the detection in dmd doesn't switch to mingw, so you have to specify the mingw-library path manually. It creates an executable depending on vcruntime140.dll and ucrtbase.dll.

If I rename vcruntime140.lib to msvcrt140.lib, the detection kicks in, but legacy_stdio_definitions.lib is not added to the linker command line automatically. If done explicitly, the resulting executable crashes.

kinke:
So in the 1st case, DMD adds `legacy_stdio_definitions.lib` automatically but apparently somehow ends up linking in the official MS one? And that one works, but our .lib results in a crash?

I have no idea how DMD sets up the environment (variables) for the linker process...

rainers:
> So in the 1st case, DMD adds `legacy_stdio_definitions.lib` automatically but apparently somehow ends up linking in the official MS one?

No, if I run this command line:
```
dmd -m32mscoff -mscrtlib=vcruntime140 args.d -L/LIBPATH:dmd2\windows\lib32mscoff\mingw -v
```
this is the linker command line:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.24.28314\bin\HostX64\x86\link.exe /NOLOGO "args.obj"   /DEFAULTLIB:phobos32mscoff /LIBPATH:dmd2\windows\lib32mscoff\mingw /OPT:NOICF  /LIBPATH:"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.24.28314\lib\x86" legacy_stdio_definitions.lib /LIBPATH:"C:\Program Files (x86)\Windows Kits\10\Lib\10.0.18362.0\ucrt\x86" /LIBPATH:"C:\Program Files (x86)\Windows Kits\10\lib\10.0.18362.0\um\x86"
```
Note that libvruntime140.lib is not in the VC lib folder (the one there has no version number). As the lib doesn't start with `msvcrt`, lld-link.exe is not invoked and legacy_stdio_definitions.lib is placed on the command line.

Running
```
dmd -m32mscoff -mscrtlib=msvcrt140 args.d -L/LIBPATH:dmd2\windows\lib32mscoff\mingw -v
```
produces thiss linker command line:
```
C:\tmp\d\dmd2\windows\bin\lld-link.exe /NOLOGO "args.obj"   /DEFAULTLIB:phobos32mscoff /LIBPATH:dmd2\windows\lib32mscoff\mingw /OPT:NOICF  /LIBPATH:"C:\tmp\d\dmd2\windows\bin\..\lib32mscoff\mingw"
```
missing legacy_stdio_definitions.lib. If added, the link succeeds, but the executable crashes. In a debugger, it seems writeln throws an exception "Attempting to write to closed File".

If I run the linker myself and replace lld-link.exe with MS link.exe, the program runs fine.

kinke:
Okay, i.e., apparently an LLD issue. I assume its v8; you could quickly recheck with recent LDC's bundled lld-link.exe (v9). Does x64 work?

rainers:
> you could quickly recheck with recent LDC's bundled lld-link.exe (v9). Does x64 work?

The new build from https://github.com/dlang/installer/pull/419 shows the same issue, also for x64.

> [I have a gut feeling this might have to do with SafeSEH.]

It rather seems to have to do with picking the wrong symbol: lld replaces __acrt_iob_func with nullfunc (from druntime's msvc.c) instead of the symbol from ucrtbase.lib.

lld-link builds a working executable if I pass ucrtbase.lib explicitly on the command line (it is searched implicitly otherwise, so maybe an ordering issue).

kinke:
> lld-link builds a working executable if I pass ucrtbase.lib explicitly on the command line (it is searched implicitly otherwise, so maybe an ordering issue).

So https://github.com/dlang/installer/blob/master/windows/mingw/msvcrt_stub.c#L37 isn't enough for LLD to search that lib for a non-weak `__acrt_iob_func` [I assume it's a regular strong symbol; can't check myself right now]? - LDC's druntime is tailored for VS 2015+, so there are no weak `__acrt_iob_func` etc., and hence no problem.

PR #415, opened 2020-01-25
Add building windows installer to azure pipelines

rainers:
to make the build process more visible and allow testing PRs to the installer

### DIFF
--- a/.azure-pipelines/build-docs.yml
+++ b/.azure-pipelines/build-docs.yml
@@ -1,0 +1,22 @@
+steps:
+  - script: |
+      export
+    displayName: Print environment variables
+  - checkout: self
+    fetchDepth: 1
+  - script: |
+      git clone --branch $(BRANCH) https://github.com/dlang/dlang.org.git work/dlang.org
+      git clone --branch $(BRANCH) https://github.com/dlang/dmd.git work/dmd
+    displayName: clone dlang.org
+  - script: |
+      cd work/dlang.org
+      make -f posix.mak release
+    displayName: Build docs
+  - script: |
+      mkdir -p work/docs
+      zip -r work/docs/docs.zip work/dlang.org/web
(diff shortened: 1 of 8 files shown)


rainers:
I've cleaned up the commit history and reenabled the other jobs (LLD/mingw).

On Appveyor, we avoided unnecessary builds of LLD and mingw using different branches. Shouldn't we do the same for azure? I'm thinking of updating the installer with 64-bit binaries, but that would need another long LLD build.

wilzbach:
> On Appveyor, we avoided unnecessary builds of LLD and mingw using different branches. Shouldn't we do the same for azure? I'm thinking of updating the installer with 64-bit binaries, but that would need another long LLD build.

I moved them back in the main branch as this repo hasn't seen much activity and I saw the risk of the build getting outdated a bit (+lost) higher than the additional build cost (especially as we got 10 parallel builds with Azure and didn't need to be so resourece-saving anymore).

Anyhow, yeah if you plan to do more work, it makes sense to move these builds back in a branch. What do you think about one branch for all the artefact-generating jobs?

rainers:
An alternative might be that the builds jobs check whether the artifact is already available on downloads.dlang.org and skip the build if no new version is going to be generated.

rainers:
> What do you think about one branch for all the artefact-generating jobs?

Maybe simpler (to avoid having to keep mutliple branches uptodate):
- master has all the latest build instructions for prerequisite packages, but disabled with `condition=false`.
- when updating a prerequisite, the PR enables the pipeline alongside changes to the build process
- the artifact of the PR is uploaded to downloads.dlang.org
- the PR is modified back to `condition=false` before being merged

wilzbach:
> An alternative might be that the builds jobs check whether the artifact is already available on downloads.dlang.org and skip the build if no new version is going to be generated.

I like this very much -> https://github.com/dlang/installer/pull/418

PR #416, opened 2020-01-25
Use script file location as default installation path.

veelo:
This makes it easier to deviate from the default (`~\dlang`). More importantly, it allows a natural Windows path, so a POSIX terminal is no longer necessary.

This change only breaks your routine if you have been using the default path for compilers, but different path for the script.


script/install.sh
```diff
@@ -119,10 +121,39 @@ fetch() {
...
+# Returns false if the script is invoked from a Windows command prompt.
+posix_terminal() {
+    # If run from a POSIX terminal (including under MSYS2 or Cygwin) the TERM
+    # variable is defined to something like "xterm". If run from a Windows
+    # command prompt through bash.exe from an MSYS installation, TERM keeps
+    # its default value, which is "cygwin".
+    if [[ $TERM == "cygwin" ]]; then
+        return 1;   # false
+    else
+        return 0;   # true
```
wilzbach: This seems sketchy and will confuse future readers.

script/install.sh
```diff
@@ -119,10 +121,39 @@ fetch() {
...
+    fi
+}
+
 COMMAND=
 COMPILER=dmd
 VERBOSITY=1
-ROOT=~/dlang
+if [ -z "$SCRIPT_PATH" ]; then
+    # We are being piped into bash from stdin. Set a default.
```
wilzbach: That's not what's being checked here.

script/install.sh
```diff
@@ -119,10 +121,39 @@ fetch() {
...
+    if posix_terminal; then
+        ROOT=~/dlang
+    else
+        # Default to a ROOT that is outside the POSIX-like environment.
+        ROOT=$(cygpath "$LOCALAPPDATA")/dlang
+    fi
+else
+    # Default to the script location.
+    if command -v cygpath &>/dev/null; then
```
wilzbach: Isn't this done similarly in other places already? Maybe time to wrap in a function?

script/install.sh
```diff
@@ -5,6 +5,8 @@
 # Authors: Martin Nowak
 # Documentation: https://dlang.org/install.html
 
+SCRIPT_PATH=${BASH_SOURCE[0]}
```
wilzbach: Let's better use the full expression as given and explained here: https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself

tl:dr: it matters for a few special cases.

veelo: We should first agree on what should happen if the script is called through a symlink. Should compilers be installed alongside the original (then the full expression is needed) or alongside the symlink (works as is)? I think the latter is desirable. You might want to have your installer git repository in `~/projects/installer`, and test it in `~/dlangtest/`. That is possible with `ln -s ~/projects/installer/script/install.sh ~/dlangtest/install.sh` and calling it through `~/dlangtest/install.sh`.

script/install.sh
veelo: Would this be better?
```
    if [[ $TERM == "cygwin" ]]; then
        return `false`;
    else
        return `true`;
    fi
```

veelo: I'll change it into
```
    if [[ $TERM == "cygwin" ]]; then
        return $(false);
    else
        return $(true);
    fi
```

script/install.sh
veelo: Clarified in comment.

veelo:
It seems to me that the build failure is not due to the PR, but something else. Correct? Will it give it another go on its own, or does it need a push? Thanks.

veelo:
Superseded by #454.

PR #417, opened 2020-01-28
build and install 64-bit dmd.exe

rainers:


create_dmd_release/create_dmd_release.d
```diff
@@ -667,16 +660,23 @@ void createRelease(string branch)
...
+            sc_ini = sc_ini.replace(`%@P%\optlink.exe`, `%@P%\..\bin\optlink.exe`);
+            std.file.write(releaseBin64Dir~"/sc.ini", sc_ini);
+        }
+        else // Win doesn't include 64-bit tools
         {
-            copyFile(cloneDir~"/dmd/generated/"~osDirName~"/release/64/dmd"~exe, releaseBin64Dir~"/dmd"~exe);
             copyDir(cloneDir~"/tools/generated/"~osDirName~"/64", releaseBin64Dir, file => !file.endsWith(obj));
             copyFile(cloneDir~"/dub/bin/dub64"~exe, releaseBin64Dir~"/dub"~exe);
-            if (codesign)
-                signBinaries(releaseBin32Dir);
```
rainers: This would have never signed the 64-bit executables, but there have been none so far. Signing should include DLLs, too, though.

wilzbach:
You might want to add a changelog entry for this as I think the windows community has been anticipating this for a while.

rainers:
> You might want to add a changelog entry

Will do, but I hope to get a couple of other changes done, too.

PR #419, opened 2020-01-28
build LLD for x64, too

rainers:
dmd expects lld-link.exe in the same directory, so build an appropriate version for the bin64 folder.


wilzbach:
http://downloads.dlang.org/other/lld-link-9.0.0-x64.zip
http://downloads.dlang.org/other/lld-link-9.0.0-x64.zip.sig

PR #420, opened 2020-01-29
build curl on azure, update to 7.68.0

rainers:


wilzbach:
Awesome!
http://downloads.dlang.org/other/libcurl-7.68.0-WinSSL-zlib-x86-x64.zip
http://downloads.dlang.org/other/libcurl-7.68.0-WinSSL-zlib-x86-x64.zip.sig

wilzbach:
Upgrade PR: https://github.com/dlang/installer/pull/421

PR #421, opened 2020-01-30
Update curl to 7.68 and LLD to 9.0.0

wilzbach:
From https://github.com/dlang/installer/pull/420 and https://github.com/dlang/installer/pull/418


wilzbach:
The dub registry is down atm which is why the build docs step failed.

PR #422, opened 2020-01-31
do not build dub with debug info

rainers:
I suspect that this isn't done deliberately, as no other tools have debug information compiled in.

The Windows build shrinks from almost 11 MB to about 3 MB. The linux binary will probably have even more debug info (binary size is currently 20 MB compared to other tools at 2-4 MB).

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -546,10 +546,22 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
         // too easily with the latest compiler, e.g. for nightlies
         info("Building Dub "~bitsDisplay);
         changeDir(cloneDir~"/dub");
-        version (Windows)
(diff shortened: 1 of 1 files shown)


wilzbach:
These scripts have recently been unified into build.d - https://github.com/dlang/dub/blob/master/build.d
So the replacements shouldn't work anymore.

rainers:
Ah, ok. So setting DFLAGS will work on Windows, too. I didn't like patching the script anyway...

I guess we should switch the installer test to use the same branch as the dependencies, it currently builds dmd and tools from stable.

wilzbach:
> I guess we should switch the installer test to use the same branch as the dependencies, it currently builds dmd and tools from stable.

Yes, I think using master for everything is best.

rainers:
> These scripts have recently been unified into build.d -

Changed it to use build.d if available, but unfortunately, the last tagged version of dub is used, and build.d is only on master.

PR #425, opened 2020-01-31
build dmd.exe with LDC

rainers:
What's not optimal is that the LDC-built version gets only limited coverage in the dmd tests, only the win32-ldc pipeline.

### DIFF
--- a/windows/build_release.bat
+++ b/windows/build_release.bat
@@ -14,6 +14,10 @@ set VCDIR=%VCToolsInstallDir%
 set SDKDIR=.
 set MSVC_AR=%VCToolsInstallDir%\bin\Hostx64\x64\lib.exe
 
+rem enable autodetection in LDC, so it doesn't mix x86/x64 libs
+set LDC_VSDIR=%VSINSTALLDIR%
+set VSINSTALLDIR=
+
 cd create_dmd_release
 
 "%HOST_DC%" -m32 -gf build_all.d common.d -version=NoVagrant || exit /B 1
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -48,6 +48,8 @@ version(Windows)
 else
     enum platforms = [linux_both, windows_both, osx_both, freebsd_32, freebsd_64];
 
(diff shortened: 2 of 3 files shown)


rainers:
A random benchmark for this build:
```
dmd.exe -unittest -c std\regex\internal\tests.d
```
- dmd 2.090.0: 10.640 sec (debug build was released)
- dmd stable:     5.610 sec (https://github.com/dlang/dmd/pull/10721 applied)
- this installer:        3.550 sec

windows/build_release.bat
```diff
@@ -14,6 +14,9 @@ set VCDIR=%VCToolsInstallDir%
 set SDKDIR=.
 set MSVC_AR=%VCToolsInstallDir%\bin\Hostx64\x64\lib.exe
 
+rem enable autodetection in LDC, so it doesn't mix x86/x64 libs
+set VSINSTALLDIR=
```
kinke: This makes LDC use LLD and the MinGW libs. Use this for proper x86/x64 setup:
```batch
set LDC_VSDIR=%VSINSTALLDIR%
set VSINSTALLDIR=
```

rainers: > This makes LDC use LLD and the MinGW libs.

Oops, indeed. Thanks for the tip.

anass-O:
30%+ increase, LGTM

kinke:
Let me be pedantic: a compile time *decrease* by almost 37% means it's 1.58x as fast, and thus 58% faster. ;) - Rainer, you might want to update to v1.20 now. Btw, I guess the Linux and Mac builds are compiled with DMD?

rainers:
rebased and upgraded LDC to 1.20.0

> Btw, I guess the Linux and Mac builds are compiled with DMD?

Yes, I'm not too familiar with the setup on these platforms.

wilzbach:
> rebased and upgraded LDC to 1.20.0

Rebased again as I merged your x64 LLD PR.

rainers:
> Rebased again as I merged your x64 LLD PR.

Thanks. Let's hope Martin doesn't have too much trouble upgrading his Windows VM to VS2015 or later.

PR #428, opened 2020-02-08
[WIP] fix Issue 20489 - Installer deleting files after install

rainers:
wait for uninstaller to terminate

As I cannot reproduce the issue locally, let's wait for the reporter to confirm it actually helps.


Biotronic:
Looks pretty good - I may have managed to recreate the issue once, but after 15 attempts I've given up on triggering it again, so I'm prepared to call that a fluke.

PR #429, opened 2020-02-08
Windows installer: "reserve" files to avoid pauses when switching between pages

rainers:
`ReserveFile` inside a section doesn't seem to work


rainers:
> @rainers you can go ahead and merge

Ok, will do.

PR #431, opened 2020-02-08
windows installer: detect VS2019 and VS2017/VS2019 Build Tools to not…

rainers:
… warn about missing VC runtime

copied from Visual D: https://github.com/dlang/visuald/blob/master/nsis/visuald.nsi#L1663

### DIFF
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -385,6 +385,26 @@ FunctionEnd
 Function DetectVC
     ClearErrors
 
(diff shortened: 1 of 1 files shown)


Geod24:
Can you confirm if this fixed https://issues.dlang.org/show_bug.cgi?id=20407 ?

rainers:
Just tried it, dmd 2.092 doesn't ask to install VS when only VS2019 is installed.

PR #438, opened 2020-02-19
azure: do not skip verifying signatures of downloaded archives

rainers:
Let's see if the signatures can be verified on azure...

### DIFF
--- a/.azure-pipelines/build-winrelease.yml
+++ b/.azure-pipelines/build-winrelease.yml
@@ -26,6 +26,8 @@ steps:
     displayName: Unzipping docs
   - script: |
       @echo on
+      powershell -command "& { iwr https://dlang.org/d-keyring.gpg -OutFile d-keyring.gpg }"
+      gpg --import d-keyring.gpg
       call windows/build_release.bat
     displayName: Build release
   - script: |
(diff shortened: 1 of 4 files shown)


rainers:
Skipping verification of lld and mingw now as no valid signatures are available, in the hope of unblocking nightliy/release builds.

.azure-pipelines/build-winrelease.yml
```diff
@@ -26,6 +26,9 @@ steps:
     displayName: Unzipping docs
   - script: |
       @echo on
+      gpg --import create_dmd_release/codesign/dlang.pgp
+      gpg --import create_dmd_release/codesign/nowak.pgp
+      gpg --import create_dmd_release/codesign/wilzbach.pgp
```
wilzbach: Uhm you can download them as a keyring: https://dlang.org/gpg_keys.html

That's how the install.sh works as well.

rainers: Downloading them from external sources seems to somewhat defeat the purpose of verifying the archives. Is it standard to do this?
I tried `--receive-keys` initially but that caused an I/O error on the keyserver.

rainers: Or do you mean that the keyring file should be added to this repo?

wilzbach: Yes, probably a better idea then though as I mentioned earlier I think sha256 hashes might be easier as then everyone with write access can upload files and not only Martin and myself.

rainers:
Now switched to sha256 for newer archives, but downloading the keyring file for simplicity (azure doesn't build official releases anyway).

PR #439, opened 2020-02-21
document changes to the Windows installation

rainers:
integrates the two separate existing changelog entries.


rainers:
Is there a preview available somewhere? I didn't find any pending.html in the generated docs.

wilzbach:
It should appear here: https://dlang.org/changelog/pending, but I think it's only redeployed for the dmd,druntime, phobos repos so I guess it should appear within the next few hours.

rainers:
Problem is that the nightlies are currently broken (due to bad signatures, hopefully fixed now) and the 64-bit build of dmd.exe (not sure if Martin already updated the VM).

PR #440, opened 2020-02-22
update to MSVC BuildTools 2017

MartinNowak:
- use vcvarsall.bat to setup correct compiler environment
  (rather than explicitly including things)
- set CC, AR, LIB, and CC32 to plain command names
  (leaving it to vcvarsall.bat to resolve them)
- BuildTools 2019 would fail to execute on Windows 7
  (might miss some updates)

### DIFF
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -318,7 +318,7 @@ void runBuild(ref Box box, string ver, bool isBranch, bool skipDocs)
                 dmd = `old-dmd\dmd2\windows\bin\dmd.exe`;
             else
                 dmd = `ldc\ldc2-`~ldcVer~`-windows-multilib\bin\ldmd2.exe`;
-            rdmd = `old-dmd\dmd2\windows\bin\rdmd.exe --compiler=`~dmd;
+            rdmd = `old-dmd\dmd2\windows\bin\rdmd.exe`;
             break;
         case OS.osx:
             dmd = "old-dmd/dmd2/osx/bin/dmd";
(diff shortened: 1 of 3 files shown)


rainers:
@MartinNowak  All the new stuff is not on stable, but master. Did you mean to target the latter instead?

MartinNowak:
Will rebase so it tests properly, just figured I need this to build the release. But you're right, master wasn't yet branched, so it should go there.

rainers:
> `create_dmd_release.d(466): Error: undefined identifier msvcBin32Dir`

I'm not sure this can really work as is: The compiler needs to be different for Win32 and x64 builds. 

Depending on the version installed for dmc, it also comes with a cl.exe that can get in the way.

@MartinNowak Have you tried master and setting up the environment as in https://github.com/dlang/installer/blob/master/windows/build_release.bat ?

rainers:
> I'm not sure this can really work as is: The compiler needs to be different for Win32 and x64 builds.

Ah, I see you are calling vcvarsall.bat in between runs, that could work.

I've removed the failing line and pushed to your repo.

MartinNowak:
This https://github.com/dlang/installer/pull/440/files#diff-5c352fb76ca5590e7ad3d2640c695272R375 and https://github.com/dlang/installer/blob/master/windows/build_release.bat#L19 colli Would you want to try the vcvarsall approach from this PR in windows_build? Seems more future-proof than activating https://github.com/dlang/installer/blob/master/windows/build_release.bat#L2 (x64) while also using the x86 compiler.
Not sure what the LDC and VCINSTALLDIR story is about. If that's a must, we could hide the envvar before/when calling LDC.

rainers:
> Not sure what the LDC and VCINSTALLDIR story is about. If that's a must, we could hide the envvar before/when calling LDC.

That's what I'm currently trying to do, but LDC is rather restive.

rainers:
Pushed a fix that worked for me locally.

rainers:
BTW: if the release is built with VS2017 or later we should also integrate https://github.com/dlang/druntime/pull/2946 because older VC runtimes are also affected.

rainers:
@MartinNowak It looks like VSINSTALLDIR is not yet set in the Windows VM started by buildkite.

PR #441, opened 2020-02-25
fixups for VS2017

MartinNowak:
- fix dub build
- fix usage of correct vsvarsall env
- fix codesign script

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -516,7 +516,7 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
         {
             // v1.20+
             version (Windows)
-                run(msvcVars~"SET DMD="~hostDMD~" && "~hostDMD~" -run build.d -O -w -m"~bitsStr);
+                run(msvcVars~"SET DMD="~hostDMD~" && "~hostDMD~" -m"~bitsStr~" -run build.d -O -w -m"~bitsStr);
             else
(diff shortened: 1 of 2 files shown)


create_dmd_release/create_dmd_release.d
```diff
@@ -471,13 +468,13 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
     {
         info("Building Druntime 32mscoff");
         changeDir(cloneDir~"/druntime");
-        run(msvcVarsX86~makecmd~msvcEnv~" druntime32mscoff");
+        run(msvcVars~makecmd~msvcEnv~" druntime32mscoff");
```
rainers: You have to switch to the x86 environment here to get the correct compiler for -m32mscoff.

MartinNowak: I think that's done by `msvcVars` which activates either x86 or x64, see https://github.com/dlang/installer/pull/441/files#diff-5c352fb76ca5590e7ad3d2640c695272R385.

MartinNowak: I see, that's a 32-bit build in the 64-bit branch, weird that it worked.

rainers: I think it will compile and create a library, but it will have mixed architecture in it (x64 object files from cl.exe and 32-bit files from dmd).

MartinNowak: Fixed now.

MartinNowak:
> Is stable correct now?

Yes

PR #442, opened 2020-02-25
fix NSIS installer warnings

MartinNowak:
- remove description for dropped dmd1 download section
- build unicode instead of deprecated ansi installer


rainers:
I have moved to Unicode setting above the includes to fix the failure.

MartinNowak:
> I have moved to Unicode setting above the includes to fix the failure.

Oh, apparently forgot to push that fixup, thanks.

PR #449, opened 2020-04-30
fix SHA for LDC 1.21.0

rainers:


rainers:
The build still fails with error
```
std.file.FileException@std\file.d(4234): Copy from C:\Users\VssAdministrator\AppData\Local\Temp\tmp.CC2C39\clones/dmd/src\.azure-pipelines/lib.sh to C:\Users\VssAdministrator\AppData\Local\Temp\tmp.CC2C39/dmd.master.windows/dmd2/src/dmd\.azure-pipelines/lib.sh: The system cannot find the path specified.
```
This file was never copied until recently (the first failing test is about a week ago). @MartinNowak Is it a deliberate change to copy all of the dmd folder now? I don't see that in the lastest beta.
I would rather go in the other direction: remove all the unnecessary sources from the binary distribution, anyone interested should get the sources from github.

veelo:
@rainers it seems to me that `git ls-files` is malfunctioning. https://issues.dlang.org/show_bug.cgi?id=20826

PR #450, opened 2020-05-12
Fix typo.

veelo:


veelo:
Thanks @thewilsonator. Frankly I made this change to test the health of CI, as #416 has been stalled because of what looks like flaky CI. Indeed CI is still flaky, although it broke at a different link in the chain this time.

I can imagine you need a green CI before you can merge #416 because that does not simply fix a typo, and I wonder what I can do to help make that happen...

I am working towards automated installation and activation of the required compiler versions on Windows (outside Cygwin) for use in CI.

PR #451, opened 2020-05-12
Do not merge

veelo:
Debugging the CI.


rainers:
Thanks for looking into this.

It seems the problem is that `git ls-files` lists all files in the project, not only the files in the current directory (anymore?). It doesn't do this for me locally (git version 2.22.0). Can you add printing the git version on the CI?

veelo:
My local git version is 2.23.0.windows.1, which does not have this problem. The CI runs 2.26.2.windows.1, which is the most recent maintained build. Will try to reproduce locally.

PR #452, opened 2020-05-13
Fix CI checks.

veelo:
`create_dmd_release` was not working with the right clones directory, causing confusing git behaviour.

### DIFF
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -197,9 +197,9 @@ private:
         return res.outdent();
     }
 
-    auto build(string ver, bool isBranch, bool skipDocs)
+    auto build(string ver, bool isBranch, bool skipDocs, string clones)
     {
-        return runBuild(this, ver, isBranch, skipDocs);
+        return runBuild(this, ver, isBranch, skipDocs, clones);
     }
 
     ~this()
@@ -288,7 +288,7 @@ void prepareExtraBins(string workDir)
 //------------------------------------------------------------------------------
 // builds a dmd.VERSION.OS.MODEL.zip on the vanilla VirtualBox image
 
-void runBuild(ref Box box, string ver, bool isBranch, bool skipDocs)
+void runBuild(ref Box box, string ver, bool isBranch, bool skipDocs, string clones)
 {
     with (box.shell())
     {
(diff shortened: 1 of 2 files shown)


veelo:
Check fails due to https://gdcproject.org being down :-(

https://forum.dlang.org/post/bwoboqzuwlctrlynqcez@forum.dlang.org

rainers:
@veelo Thanks for figuring this out. It seems the commit history contains extra old commits, please rebase.

@MartinNowak Do you see any issues with this?

veelo:
> @veelo Thanks for figuring this out.

Glad I could help.

> It seems the commit history contains extra old commits, please rebase.

This is where my git proficiency falls short... The master branch of my clone on github has extra commits due to merging in changes from upstream, and my clone at home has too. I see no way to remove these, and I can only rebase on my own branches. I synchronize my github clone with upstream by creating a PR through the github API, and that leaves a trail. Am I doing something wrong? How can I synchronize my master branch without introducing noise? And how do I rebase this PR on the upstream master to remove that noise?

(I always seem to run into this, and if this is covered in git 101 at all, it is outside of my attention span. The problem with git IMO is that you have to be a git professional to be able to do what you want to do, in addition to the professions that really matter.)

MoonlightSentinel:
You can either delete `master` and re-track `upstream/master` 
```
git checkout <other branch>
git branch -D master
git checkout -t upstream/master
git push --force --set-upstream origin HEAD
```
or remove these commits using an interactive rebase:
```
git checkout master
git rebase -i upstream/master
git push --force
```
That should give you a list of all additional commits which you need to drop (see the description below).

Once you have restored your master branch the following should do the trick:
```
git checkout <PR branch>
git rebase -i HEAD~<Amount of commits in your PR> master
git push --force
```

veelo:
Wow @MoonlightSentinel I guess I'll need to print this out and frame it in! Thanks!!

veelo:
Can someone please merge this?

MartinNowak:
I don't quite see what the original problem was? This now escapes the *local* path to the clones directory, but passes it to a command run on the *remote* virtual machine.

The copy commands still reference the relative `clones` folder. Can't we just stick with that?
https://github.com/dlang/installer/blob/ff600fc97089606ae94ee1310b7540c417731229/create_dmd_release/build_all.d#L309-L310

PR #454, opened 2020-05-15
Support install.sh being called from Windows cmd.

veelo:
The `install.sh` compiler version manager no longer requires running a POSIX terminal emulator. When run from the Windows command prompt it installs compilers in `%USERPROFILE%\dlang` and provides `activate.bat` for amending `PATH`.

### DIFF
--- a/changelog/install_sh_windows.dd
+++ b/changelog/install_sh_windows.dd
@@ -1,0 +1,29 @@
+The install script now also works on the Windows command prompt.
+
+The official D version manager ($(LINK2 https://dlang.org/install.sh, install.sh),
+documented at $(LINK2 https://dlang.org/install.html, dlang.org/install.html))
+used to require a POSIX environment. On Windows, prior to this release, this
+meant running the script and the compilers that it installed from within a
+POSIX terminal emulator as provided by MSYS2 or Cygwin. With this release it is
+now possible to install and activate compilers from the Windows command prompt
+directly.
+
+Assuming an MSYS2 installation is present in `C:\msys64`, necessary decompression
+tools can be installed from the Windows command prompt:
+-------
+C:\msys64\usr\bin\pacman.exe --sync unzip p7zip
+-------
+
+The following will invoke the script directly from dlang.org and install the
+latest dmd compiler in the `%USERPROFILE%\dlang` folder:
+-------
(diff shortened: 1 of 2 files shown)


veelo:
Thanks!

Geod24:
That's amazing, thanks @veelo ! Could you also raise a PR against `dlang.org`'s download page ?
https://github.com/dlang/dlang.org/blob/a682f4be14004a9e213d003874186302f9a3cb10/download.dd#L91

veelo:
Thank you @Geod24. Looking at https://dlang.org/download.html, I have difficulty to see a nice way of documenting this here without cluttering up the page. Invocation of the script on Windows differs a bit from POSIX (and is longer), and requires different precursory steps depending on Bash being used from Git or MSYS2. This is documented on https://dlang.org/install.html (oof I now see that the master branch is used for production -- so the functionality described there is not released yet!). Maybe just adding a button so you'll see [exe][7z][script] is the best? A mention on https://dlang.org/dmd-windows.html would probably be in place, though.

Note that both Git and MSYS2 come with curl as well, so running the script directly off dlang.org is possible, but the security of executing unseen unsigned scripts from some website is questionable, so I'm wary of promoting this.

My evil plan was originally to use the script to automatically install a compiler that satisfies any `toolchainRequirements` in `dub.json`. I'm now leaning more towards integrating this functionality into dub directly, as it is already in the business of downloading, signature checking and installing. This would also remove the need for bash (first directive in [secure shell programming](https://sipb.mit.edu/doc/safe-shell/): don't use shell).

Geod24:
> Note that both Git and MSYS2 come with curl as well, so running the script directly off dlang.org is possible, but the security of executing unseen unsigned scripts from some website is questionable, so I'm wary of promoting this.

A reasonable concern, however this boat has sailed a long time ago, since `install.sh` has been there for years. 

> (oof I now see that the master branch is used for production -- so the functionality described there is not released yet!)

Yep that isn't great, to say the least. I was wondering if, the script being on `dlang.org`, it would be available immediately, but that doesn't seem to be the case. You could raise the PR as a draft to be merged once a new release is out / the script is deployed.

> (first directive in secure shell programming: don't use shell).

Yeah, use D!

veelo:
> Yeah, use D!

https://github.com/dlang/dub/issues/1954

PR #457, opened 2020-06-13
add auto-update for the d-keyring

AntonOks:
@wilzbach based on the https://github.com/dlang/dlang.org/pull/2817 discussion

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -410,8 +410,9 @@ parse_args() {
 # run_command command [compiler]
 run_command() {
     # check if tools where installed already and update d-keyring.gpg if necessary
-    local tmp=$(mkdtemp)
-    if [ -f "$ROOT/d-keyring.gpg" ] && [ "$(command curl https://dlang.org/d-keyring.gpg -z $ROOT/d-keyring.gpg -o $tmp/d-keyring.gpg -s -L -w %{http_code})" == "200" ] ; then
+    local tmp
+    tmp=$(mkdtemp)
+    if [ -f "$ROOT/d-keyring.gpg" ] && [ $(command curl https://dlang.org/d-keyring.gpg -z "$ROOT/d-keyring.gpg" -o "$tmp/d-keyring.gpg" -s -L -w "%{http_code}") == "200" ] ; then
         mv "$tmp/d-keyring.gpg" "$ROOT/d-keyring.gpg"
         log "auto-updated $ROOT/d-keyring.gpg"
     fi
--- a/script/install.sh
+++ b/script/install.sh
@@ -5,6 +5,8 @@
 # Authors: Martin Nowak
 # Documentation: https://dlang.org/install.html
 
+# set -x
+
 _() {
 
 # Returns false if the script is invoked from a Windows command prompt.
@@ -407,6 +409,13 @@ parse_args() {
 
 # run_command command [compiler]
 run_command() {
+    # check if tools where installed already and update d-keyring.gpg if necessary
+    local tmp=$(mkdtemp)
+    if [ -f "$ROOT/d-keyring.gpg" ] && [ "$(command curl https://dlang.org/d-keyring.gpg -z $ROOT/d-keyring.gpg -o $tmp/d-keyring.gpg -s -L -w %{http_code})" == "200" ] ; then
+        mv "$tmp/d-keyring.gpg" "$ROOT/d-keyring.gpg"
+        log "auto-updated $ROOT/d-keyring.gpg"
+    fi
+
     case $1 in
         install)
             check_tools curl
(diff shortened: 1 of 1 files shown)


AntonOks:
@wilzbach @MartinNowak Hi there. Seems like all checks are ok finally. Is there anything else needed or desired or are you also ok with this?

thanks and regards

AntonOks:
3 weeks passed by... any feedback from anyone would be appreciated...

thewilsonator (approved):

script/install.sh
```diff
@@ -5,6 +5,8 @@
 # Authors: Martin Nowak
 # Documentation: https://dlang.org/install.html
 
+# set -x
```
wilzbach, line 8: Shouldn't have been committed

script/install.sh
```diff
@@ -80,7 +82,7 @@ retry() {
     for i in {0..4}; do
         if "$@"; then
             break
-        elif [ $i -lt 4 ]; then
+        elif [ "$i" -lt 4 ]; then
```
wilzbach, line 85: `i` is a number, not a string.

script/install.sh
```diff
@@ -407,6 +409,14 @@ parse_args() {
 
 # run_command command [compiler]
 run_command() {
+    # check if tools where installed already and update d-keyring.gpg if necessary
+    local tmp
+    tmp=$(mkdtemp)
+    if [ -f "$ROOT/d-keyring.gpg" ] && [ "$(command curl https://dlang.org/d-keyring.gpg -z "$ROOT/d-keyring.gpg" -o "$tmp/d-keyring.gpg" -s -L -w "%{http_code}")" == "200" ] ; then
```
wilzbach, line 415: This is the real problem:
1) This probably needs retry logic
2) Probably needs the fallback mirrors
3) Should verify the authenticity
4) Shouldn't re-download the file(s) if they haven't changed
5) Needs to update `install.sh` as well (the script changes a lot more often than the keyring)
6) Needs to provide the user an option to opt-out
7) Should probably use "Last-Modified", "ETag" or similar to avoid unnecessary downloads
8) Auto-updates need to be documented in help page, documentation and changelog (there are serious security concerns with it)

wilzbach:
Hi, I'm really sorry for not responding (things are rather busy), but this shouldn't have been merged without review. I triggered a revert, s.t. this change doesn't get accidentally deployed before we are ready to announce it. I am aware that this is a very frustrating experience and I will try to be a lot more active next time and if you understandably decide to not resubmit, I will try to find time to champion it. Sorry.

script/install.sh
AntonOks: forgot this after "debugging", but then again, it is commented out and shouldn't harm, that's why I finally decided to not delete it again. anyhow, feel free to remove it...

script/install.sh
AntonOks: correct.
if I'm nor wrong, I had it like you suggest it now initially but "shellcheck" complained so I made it as it is... also here, feel free to change it to whatever fit's your standards / requirements...

script/install.sh
AntonOks: this all together is far beyond my bash foo and also the time I can / want spend on this topic, sorry...
nevertheless, some comments from here...

4. => make's sense but I couldn't find a (smart & easy) way how to check if the file was changed without downloading it. actually I did read plenty of "curl way" post / blogs, i.e. https://blog.yjl.im/2012/03/downloading-only-when-modified-using.html and https://www.thegeekstuff.com/2012/04/curl-examples/ but at the end it seems like curl ALLWAYS downloads the file to be able to check for a change... and as it's downloaded anyhow, I decided to implement it as it is

5. => well, if the script itself should also be updated, neither this change nor the "update" option of the script makes any sense... because in other words you ask to make the already existing"update" option an always applied defauld (from my point of view, I wouldn't like this kind of change... I want to know and control when and why a tool gets updated)

7. => see my comments on topic nr4

AntonOks:
> 
> 
> Hi, I'm really sorry for not responding (things are rather busy), but this shouldn't have been merged without review. I triggered a revert, s.t. this change doesn't get accidentally deployed before we are ready to announce it. I am aware that this is a very frustrating experience and I will try to be a lot more active next time and if you understandably decide to not resubmit, I will try to find time to champion it. Sorry.

Hey Seb, no prob and no need for apologias.
As people wrote so many time in the forum, Dlang is only community driven anyhow... and I know you and a little group of "key contributer's" improved Dlang and it's eco system A LOT... so big thanks for all your efforts!

andre2007:
@AntonOks 
What do you think about following interaction flow:
In case of an key ring error there is a prompt for updating the file.
In addition there is also a command line argument to force the key ring update (false by default).

PR #461, opened 2020-06-23
merge stable

MartinNowak:
- remove broken and unmaintained OSX-32 builds from releases
- fix git ls-files on Windows CI machine
- Revert "Merge pull request #452 from veelo/fix_CI"


create_dmd_release/build_all.d
```diff
@@ -321,7 +325,7 @@ void runBuild(ref Box box, string ver, bool isBranch, bool skipDocs, string clon
             break;
         }
 
-        auto build = rdmd~" -g create_dmd_release --extras=extraBins --use-clone="~escapeShellFileName(clones)~" --host-dmd="~dmd;
```
rainers, line 324: IIRC @veelo fought long with this workaround, but yours seems to work, too.

PR #462, opened 2020-06-27
Make the activate generated script reentrant

andralex:
Currently sourcing `activate` several times in a row (either for the same version or different versions) messes things up. With this the `deactivate` function clears the current activation before entering the new one.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -903,7 +903,7 @@ write_env_vars() {
     echo "    unset -f deactivate"
     echo "}"
     echo
-    echo "[[ -v _OLD_D_PATH ]] && deactivate"
+    echo "if [ -v _OLD_D_PATH ] ; then deactivate; fi"
     echo "_OLD_D_PATH=\"\${PATH:-}\""
 
     if [ -n "$libpath" ] ; then
--- a/script/install.sh
+++ b/script/install.sh
@@ -903,6 +903,7 @@ write_env_vars() {
     echo "    unset -f deactivate"
     echo "}"
     echo
+    echo "[[ -v _OLD_D_PATH ]] && deactivate"
     echo "_OLD_D_PATH=\"\${PATH:-}\""
 
     if [ -n "$libpath" ] ; then


andralex:
cc @John-Colvin @MartinNowak

wilzbach:
Isn't this done on purpose? You can temporarily activate a new compiler and then deactivation will return your previous compiler.
This change would break this behaviour. I'm not sure how often this is used, but I used it myself every now and then.

andralex:
> Isn't this done on purpose? You can temporarily activate a new compiler and then deactivation will return your previous compiler.
> This change would break this behaviour. I'm not sure how often this is used, but I used it myself every now and then.

There's no breakage. The only change is that if you run activate twice it will work.

wilzbach:
>  The only change is that if you run activate twice it will work. (For the same version or for different versions.)

Currently this works:

```
. $(~/dlang/install.sh --activate dmd)
...
. $(~/dlang/install.sh --activate ldc)
...
deactivate
# back to dmd
```

With your change, this will stop to work.

andralex:
> > The only change is that if you run activate twice it will work. (For the same version or for different versions.)
> 
> Currently this works:
> 
> ```
> . $(~/dlang/install.sh --activate dmd)
> ...
> . $(~/dlang/install.sh --activate ldc)
> ...
> deactivate
> # back to dmd
> ```

It does not work because there is only one copy of the `_OLD_D*` variables. After the second activation you won't be able to deactivate the first.

script/install.sh
```diff
@@ -903,6 +903,7 @@ write_env_vars() {
     echo "    unset -f deactivate"
     echo "}"
     echo
+    echo "[[ -v _OLD_D_PATH ]] && deactivate"
```
wilzbach, line 906: Looks like this is needed for OSX:

```suggestion
    echo "if [ -v _OLD_D_PATH ] ; then deactivate; fi"
```

andralex: Works on my osx, but better safe than sorry.

PR #463, opened 2020-07-20
Fixes issue 21059: installer fails when piped on Posix.

veelo:
The assumption that $SHLVL increases when the script is run does not always hold. $SHLVL does not increase if the script is piped into bash, thereby failing to detect Posix. Also, if the default terminal shell is not bash, $SHLVL might be 1 on Posix.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -9,9 +9,11 @@ _() {
 
 # Returns false if the script is invoked from a Windows command prompt.
 posix_terminal() {
-    # If this script is run on Windows cmd by passing it as argument to bash.exe,
-    # the shell level will be 1. If it is run from a POSIX terminal, it will be > 1.
-    if [ "$SHLVL" = 1 ]; then
+    # If run from a POSIX terminal (including under MSYS2 or Cygwin) the TERM
+    # variable is defined to something like "xterm". If run from a Windows
+    # command prompt through bash.exe from an MSYS installation, TERM keeps
+    # its default value, which is "cygwin".
+    if [[ $TERM == "cygwin" ]]; then
         false
     else
         true
--- a/script/install.sh
+++ b/script/install.sh
@@ -13,7 +13,7 @@ posix_terminal() {
     # variable is defined to something like "xterm". If run from a Windows
     # command prompt through bash.exe from an MSYS installation, TERM keeps
     # its default value, which is "cygwin".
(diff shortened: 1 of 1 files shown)


jmh530:
Thanks for working on this.

Is this tested with Windows Subsystem for Linux (WSL)?

veelo:
> Is this tested with Windows Subsystem for Linux (WSL)?

No it is not tested with WSL. The installer is expected to install the Linux version of dmd on WSL, producing Linux executables (that require WSL to run on Windows).

nexor:
I've checked it locally and it seems that [issue with docker](https://github.com/dlang/installer/commit/e084815001390538bbd6fe5be7c2d4d81ee681b7#r40379696) is resolved. 
Thank you!

veelo:
Thanks for confirming, and sorry for the breakage!

script/install.sh
```diff
@@ -9,9 +9,11 @@ _() {
...
 # Returns false if the script is invoked from a Windows command prompt.
 posix_terminal() {
-    # If this script is run on Windows cmd by passing it as argument to bash.exe,
-    # the shell level will be 1. If it is run from a POSIX terminal, it will be > 1.
-    if [ "$SHLVL" = 1 ]; then
+    # If run from a POSIX terminal (including under MSYS2 or Cygwin) the TERM
+    # variable is defined to something like "xterm". If run from a Windows
+    # command prompt through bash.exe from an MSYS installation, TERM keeps
+    # its default value, which is "cygwin".
+    if [[ $TERM == "cygwin" ]]; then
```
wilzbach, line 16: Provide a default fallback:

```suggestion
    if [[ "${TERM:-noterm}" == "cygwin" ]]; then
```

script/install.sh
```diff
@@ -9,9 +9,11 @@ _() {
...
         false
```
wilzbach, line 17: there are no booleans in bash!

```suggestion
        return 0
```

veelo: But you got it backwards. `return 0` == `true` in bash, unlike D. https://stackoverflow.com/questions/23320150/bash-boolean-testing/23324368#23324368. I believe I had it this way in my first iteration and you commented rightfully that it was confusing. Good or bad style, at least this cannot be misunderstood.

wilzbach:
(deployed)

veelo:
Thanks.

PR #467, opened 2020-09-03
Revert "add auto-update for the d-keyring (#457)"

wilzbach:
This reverts commit 41a1da1de15ea6250b88ef953ff5ff2891164986.


wilzbach:
Reverting this change as we auto-update is a serious security concern and we need to do it properly and announce this change publicly.

PR #468, opened 2020-09-04
Windows: Rename VSINSTALLDIR to LDC_VSDIR

kinke:


windows/build_release.bat
```diff
@@ -1,13 +0,0 @@
-setlocal
```
rainers, line 1: I usually use this batch for local testing. Is there a way to execute (part of) the azure-scripts instead?

rainers (approved):
Otherwise LGTM

windows/build_release.bat
AndrewEdwards: I'm not sure about that. Let me defer to the professionals. @kinke, @thewilsonator, @CyberShadow. Can you provide a response to this question? Thanks.

kinke: > Is there a way to execute (part of) the azure-scripts instead?

Not that I know of.

I found it better to have all logic in the YAML, where the directories for PATH are set up appropriately (e.g., I had no idea that the `dm` dir is supposed to be the DMC dir augmented by implib.exe).

kinke:
Pinging @MartinNowak, as the Vagrant/VM setup will most likely have to be adapted.

MartinNowak:
Done https://gist.github.com/MartinNowak/8270666/revisions#diff-de5ae8181f71cdfb109e9e8d7fa69883

PR #470, opened 2020-09-06
Issue 21226 - Support installing older DMD versions

wilzbach:
I uploaded signatures files until 2.060. I doubt anyone is using older DMD installations on their CI.
After all, it even fails to run here on TravisCI.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -549,8 +549,10 @@ resolve_latest() {
             # dmd-nightly, dmd-master, dmd-branch
             # but not: dmd-2016-10-19 or dmd-branch-2016-10-20
             #          dmd-2.068.0 or dmd-2.068.2-5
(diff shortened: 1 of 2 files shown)


script/install.sh
```diff
@@ -787,10 +789,10 @@ download_and_unpack() {
         if [[ "${#files[@]}" -eq 1 && -d "${files[0]}" ]]; then
             # Single directory at the top of the archive,
             # as is common with .tar archives. Move it out.
-            find "$tmp"/target -mindepth 2 -maxdepth 2 -exec mv -t "$tmp" {} \+
```
wilzbach, line 790: `mv -t` isn't available on OSX.

wilzbach:
CC @thewilsonator

thewilsonator (approved):

PR #476, opened 2020-11-27
Fix Issue 21433 - "bash: line 952: --list-keys: command not found" wh…

CyberShadow:
…en running install.sh on Catalina

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -946,7 +946,7 @@ verify() {
     path="$1"
     urls=("${@:2}")
     : "${GPG:=$(find_gpg)}"
-    if [ "$GPG" = x ]; then
+    if [ -z "$GPG" ]; then
         return
     fi
     if ! $GPG --list-keys >/dev/null; then


script/install.sh
```diff
@@ -946,7 +946,7 @@ verify() {
     path="$1"
     urls=("${@:2}")
     : "${GPG:=$(find_gpg)}"
-    if [ "$GPG" = x ]; then
+    if [ -z "$GPG" ]; then
```
PetarKirov, line 949: I have no idea what `[ "$GPG" = x ]` was supposed to check for... perhaps the intention was to perform this check instead: `[ "$GPG" != gpg ] && [ "$GPG" != gpg2 ]`?

PetarKirov (approved):
Anyway, LGTM

script/install.sh
CyberShadow: I think it was supposed to be `[ x$GPG = x ]`, which I think is an antiquated way to check for an empty variable.

Geod24 (approved):

PR #477, opened 2020-11-30
Fix issue 21439: install.sh is disabled for LDC on FreeBSD

jacob-carlborg:


thewilsonator (approved):

jacob-carlborg:
@MartinNowak can you please deploy this?

wilzbach:
Sorry for the slow reaction time to this, but I don't think this is a good idea. If FreeBSD is supported, then it should be added explicitly and with a good error message. Without this explicit check, people using it on other systems will get all kind of hard to understand errors.

PR #478, opened 2020-12-29
Upgrade versions used by the test-suite

Geod24:
I was trying to upgrade this repository to use Github CI instead of Travis, and ran into some issues on OSX, so figured it'd be a good time to upgrade.


test/t/base.sh
```diff
@@ -3,30 +3,30 @@
 set -uexo pipefail
 
 compilers=(
-    dmd-2.064
```
Geod24, line 6: Hum, this was used to test that the old pattern (`dmd-2.064` as opposed to `dmd-2.064.0`) works, I suppose ?
Could we do it in a way that doesn't run the compiler though ?

PR #487, opened 2021-09-29
workaround OSX10.8/VSCode incompatibility with 1.27.x

MartinNowak:
- getting segfaults since ldc-1.27.0 on older release image, seemingly due to outdated build tools
- OSX release image to be updated soon, but I don't want to further delay the release


kinke:
Hmm, might be the bump to LLVM 12 with LDC v1.27, but I haven't heard of any similar reports so far. The temporary workaround seems fine (no comment/TODO, so hopefully really short-lived).

PR #488, opened 2021-11-24
Fix for LDC arm64 build

lionello:
This fixes the installer script for Apple M1 MacBooks, fixing issue https://issues.dlang.org/show_bug.cgi?id=22078.

Also see https://github.com/dlang/tools/pull/441 for a similar fix.

PR #491, opened 2021-12-28
Improve install.sh on macOS with arm64 architecture.

JohanEngelen:
See also:
https://issues.dlang.org/show_bug.cgi?id=22078
https://github.com/dlang/installer/pull/488

@lionello @maxhaton @jacob-carlborg

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -1306,14 +1306,14 @@ fi
 if [[ $OS-$ARCH = osx-aarch64 ]]; then
     if [[ $COMPILER == dmd* ]]; then
         log "
-DMD does not have builds for macOS on aarch64/arm64 achitecture (e.g. for the M1 CPU).
+DMD does not have builds for macOS on aarch64/arm64 architecture.
 Switching to x86_64 architecture (requires Rosetta).
 "
         ARCH=x86_64
     fi
     if [[ $COMPILER == ldc* ]]; then
         log "
-LDC has builds for macOS on aarch64/arm64 achitecture (e.g. the M1 CPU) since ldc-1.25.0.
+LDC has builds for macOS on aarch64/arm64 architecture since ldc-1.25.0.
 If you are installing an earlier version and get a download error,
 try '--arch x86_64' to install the x86_64 version instead (requires Rosetta).
 "
--- a/script/install.sh
+++ b/script/install.sh
@@ -400,6 +400,11 @@ parse_args() {
                 _installAction="activate"
                 ;;
 
+            --arch)
+                shift
+                ARCH_OVERRIDE=$1;
+                ;;
+
             --dmd)
                 if [ -n "$_getPathAction" ]; then
                     fatal "$1 conflicts with --${_getPathAction}"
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

thewilsonator:
Should this go to stable?

JohanEngelen:
> Should this go to stable?

Please make that a separate decision.

script/install.sh
```diff
@@ -1292,11 +1297,28 @@ case $(uname -m) in
...
+Switching to x86_64 architecture (requires Rosetta).
+"
+        ARCH=x86_64
+    fi
+    if [[ $COMPILER == ldc* ]]; then
+        log "
+LDC has builds for macOS on aarch64/arm64 achitecture (e.g. the M1 CPU) since ldc-1.25.0.
+If you are installing an earlier version and get a download error,
+try '--arch x86_64' to install the x86_64 version instead (requires Rosetta).
+"
```
jacob-carlborg, line 1319: Wouldn't it be better to detect if a version earlier than 1.25.0 is being installed?

script/install.sh
```diff
@@ -1292,11 +1297,28 @@ case $(uname -m) in
...
+fi
 if [[ $OS-$ARCH = windows-x86_64 && $COMPILER = ldc* ]]; then
     ARCH=x64
 fi
-if [[ $OS-$ARCH = osx-aarch64 && $COMPILER = ldc* ]]; then
-    ARCH=arm64
+if [[ $OS-$ARCH = osx-aarch64 ]]; then
+    if [[ $COMPILER == dmd* ]]; then
+        log "
+DMD does not have builds for macOS on aarch64/arm64 achitecture (e.g. for the M1 CPU).
```
jacob-carlborg, line 1309: I would recommend against mentioning `M1`. There are already other CPUs than M1.

script/install.sh
JohanEngelen: Yes, but version detection is work, and will not be possible for git hashes. So I decided to keep it simple, and let the user do the smart stuff ;)

script/install.sh
JohanEngelen: removed it

script/install.sh
```diff
@@ -1292,11 +1297,28 @@ case $(uname -m) in
...
 fi
-if [[ $OS-$ARCH = osx-aarch64 && $COMPILER = ldc* ]]; then
-    ARCH=arm64
+if [[ $OS-$ARCH = osx-aarch64 ]]; then
+    if [[ $COMPILER == dmd* ]]; then
+        log "
+DMD does not have builds for macOS on aarch64/arm64 architecture.
+Switching to x86_64 architecture (requires Rosetta).
+"
+        ARCH=x86_64
```
lionello, line 1312: Placing this new logic here will miss the logic on [line 1304](https://github.com/dlang/installer/pull/491/files#diff-08b624c4cf3301c03903fe40d86af08880f095510e2858cf6b0fb3cf2a71c770R1304) which makes sure we use `x64` for LDC downloads.

lionello (changes requested):

script/install.sh
JohanEngelen: Note that line 1312 is only for DMD. And it is only for OS=osx.  Using `x64` for ldc is only relevant for windows.

lionello: My bad!

lionello (approved):

script/install.sh
maxhaton: There are others but there are no others worth buying, so I think we should try to be somewhat welcoming to M1 users in this regard - probably not here but we could use a "how to use D on apple silicon" page on dlang.org since it apparently isn't obvious.

maxhaton (approved):
LGTM

script/install.sh
jacob-carlborg: Fair enough.

script/install.sh
jacob-carlborg: > There are others but there are no others worth buying,

What's wrong with M1 Pro and M1 Max? Too expensive?

> so I think we should try to be somewhat welcoming to M1 users in this regard

In general I'm against selecting out specific CPU models. I don't see any mentioning of specific Intel, AMD or Via CPU models. I've seen other open source projects that provide binaries for "Intel" or for "M1". But M1 is not an architecture, it's a CPU model. If the user has M1 Max, does that count as a M1 binary? What about the next CPU model, that would most likely be called M2? The official name of the architecture that Apple uses when targeting end users is "Apple Silicon". When targeting developers, they use "ARM64". Anything else is just going to cause confusion. I don't want the D community to do the same misstake. So far, LDC has been using "ARM64", which is perfectly fine since it's targeting developers.

> probably not here but we could use a "how to use D on apple silicon" page on dlang.org since it apparently isn't obvious.

That's a good idea, but as you said, not here.

jacob-carlborg (approved):

maxhaton:
Should've changed it to say apple silicon before I posted. if you say M1 to
me I think of the firestorm and IceStorm microarchitectures (i.e. M1 Pro
and such are an iteration of the same basic design on the CPU side at
least) rather than the actual SoC - the non-apple alternatives in the same
performance bracket are basically either some really niche arm servers,
some less niche neoverse based servers, or the neoverse devkit.

One day we might have this same conversation/thread about RISC-V but I'm
bearish on that actually happening.

JohanEngelen:
(I don't have merge rights btw)

jacob-carlborg:
> if you say M1 to me I think of the firestorm and IceStorm microarchitectures

I see. This was in the context of Mac/Apple. So I assumed CPUs made by Apple.

PR #492, opened 2022-01-13
Fix Issue 18362 - Try to enable lto for all dmd builds

maxhaton:
LTO isn't always plain sailing but let's try it.

image

A box & whisker plot of phobos test suite build times vs. flags

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -371,6 +371,8 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
     }
     auto hostDMDEnv = " HOST_DC="~hostDMD~" HOST_DMD="~hostDMD;
     auto isRelease = " ENABLE_RELEASE=1";
+    //Enable lto for everything.
+    auto ltoOption = " ENABLE_LTO=1";
     auto latest = " LATEST="~branch;
     // PIC libraries on amd64 for PIE-by-default distributions, see Bugzilla 16794
     version (linux)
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

create_dmd_release/create_dmd_release.d
```diff
@@ -371,6 +371,8 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
     }
     auto hostDMDEnv = " HOST_DC="~hostDMD~" HOST_DMD="~hostDMD;
     auto isRelease = " ENABLE_RELEASE=1";
+    //Enable lto for everything.
+    auto ltoOption = " ENABLE_LTO=1";
```
kinke, line 375: @maxhaton: Are you sure this actually works? I don't see `ENABLE_LTO` being forwarded to `build.d`in https://github.com/dlang/dmd/blob/master/src/posix.mak#L100.

maxhaton:
It's how I recorded the data. I think no one bothered to add it since it
makes the build extremely slow

create_dmd_release/create_dmd_release.d
MoonlightSentinel: GNU make sets `ENABLE_LTO=1` as an environment variable which is read by `build.d`.
So `make -f posix.mak ENABLE_LTO=1` works as expected.

kinke: Hmm okay, but what about Windows with DigitalMars make? IIRC, there's a difference, and that's why all (?) other vars are forwarded explicitly. No `ENABLE_LTO` forwarding in https://github.com/dlang/dmd/blob/master/src/win32.mak#L75 either.

kinke: => https://github.com/dlang/dmd/pull/13524

carun:
Hey @maxhaton, how did you generate and plot the graph? Do you use D for that as well? It would be interesting if you could capture the process.

PR #494, opened 2022-02-04
Add a template to build releases on Github Actions

MoonlightSentinel:
This action is based off the dmd nightlies but also includes some required changes to make it available as a reuseable workflow. This allows actual workflows to call this template, e.g. for pull requests against this repository.

Also includes a test action that uses the new action & verifies the packaged compiler by building and running some basic program.

Note: The windows checks are partially disabled due to an upstream issue
 (invalid `sc.ini` due to the Win32 OMF => MSCOFF change).

Includes the following changes to restore the currently broken release build:
- Disables LTO for FreeBSD because the generated DMD segfaults immediatly
- Provides custom DFLAGS when building the tools because the defaults found in `posix.mak` assume a freshly built dmd regardless of the actual `DMD`.

Example: https://github.com/MoonlightSentinel/installer/actions/runs/1802351875

CC @maxhaton @thewilsonator

PR #495, opened 2022-02-05
Retain original sc.ini instead of generating a new one

MoonlightSentinel:
Only append the required `LIB` path s.t. the remaining values match the expected defaults from dmd. Required because the generated config contains an invalid `LINKCMD` with the new default for 32 bit windows.

Also changes the `tools` build to use host compiler to avoid issues w.r.t. the OMF => MsCoff default transition (as LDC defaults to MsCoff for 32).

---

Extracted from #494


thewilsonator (approved):

thewilsonator:
Please merge this when you're happy with it.

MoonlightSentinel:
@thewilsonator Done, the windows build is passing again. Don't have write access to this repo though.

PR #497, opened 2022-02-08
Enforce DMD as the default compiler for rdmd

MoonlightSentinel:
Defining `-version=DefaultCompiler_DMD` ensures that rdmd defaults to DMD instead of the current host compiler (LDC).

See dlang/tools#442

CC @CyberShadow @rikkimax

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -478,7 +478,7 @@ void buildAll(Bits bits, string branch, bool dmdOnly=false)
 
         // Override DFLAGS because we're using the host compiler rather than
         // the freshly built one (posix.mak defaults to the generated dmd)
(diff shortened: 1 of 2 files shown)


rikkimax:
We need to hear from @kinke and @ibuclaw about how this relates to LDC and GDC.

Just to confirm that this really is dmd only (plus they will need to update their scripts).

kinke:
Doesn't relate at all to LDC AFAICT, LGTM.

MoonlightSentinel:
The script is explicitly tailored to build `dlang/{dmd/druntime/...}` with the corresponding build instructions. It wouldn't work for either LDC / GDC without massive changes AFAICT

rikkimax:
I've pinged @ibuclaw multiple times and it looks like next release might be coming up. LGTM.

thewilsonator (approved):

PR #496, opened 2022-02-08
Remove workaround for the broken sc.ini on Windows

MoonlightSentinel:
The configuration file was fixed upstream by dlang/dmd#13611

PR #498, opened 2022-02-11
Explicitly specify repo + branch when cloning installer

MoonlightSentinel:
`actions/checkout` defaults to the repository + branch that triggered the build, not where the action is defined. They are not set explicitly based on workflow parameters s.t. the defaults can be override for e.g. pull requests.

---

Noticed the problem when trying to redirect the nightly build to `build_release_template` - the build failed because it cloned `dmd` instead of `installer`.

### DIFF
--- a/.github/workflows/build_release_template.yml
+++ b/.github/workflows/build_release_template.yml
@@ -20,6 +20,16 @@ on:
         required: true
         description: Revision to build as a new release
 
+      installer_repo:
+        type: string
+        default: 'dlang/installer'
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

MoonlightSentinel:
Interesting, The  GA build didn't fail as observed in #496

MoonlightSentinel:
Example: [This run](https://github.com/MoonlightSentinel/dmd/runs/5158447112?check_suite_focus=true) passed using the changes from this PR.

thewilsonator (approved):

PR #499, opened 2022-02-11
build_release_template: Use windows-2019 instead of windows-latest

MoonlightSentinel:
The build script currently requires an existing VS 2019 installation. But `windows-latest` allows for runners with `windows-2022` which have VS 2022 installed.

---

Not sure why runs from my fork consistently used `windows-2019` runners.


MoonlightSentinel:
Seems like the `auto-merge` currently doesn't work in this repo?

PR #505, opened 2022-03-29
Fix Issue 22956 - Bump LLD version

maxhaton:
Or try to bump version at least

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,7 +11,7 @@ jobs:
     variables:
       VSINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\
       # NB: update windows/build_lld.sha256sums as well
-      LLVM_VER: 9.0.0
+      LLVM_VER: 14.0.0
       ARCH: x86
     steps:
       - template: .azure-pipelines/build-lld.yml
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     variables:
       VSINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\
       # NB: update windows/build_lld.sha256sums as well
-      LLVM_VER: 9.0.0
+      LLVM_VER: 14.0.0
```
kinke, line 14: Note the comment above - the checksum will need to be updated too at least.

MoonlightSentinel:
#506

maxhaton:
OK so LLVM don't do releases as the script expects anymore.

https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/lld-14.0.0.src.tar.xz is the new place we need to look apparently.

maxhaton:
I guess we should move to using the `.sig`s instead

adamdruppe:
so what's left here? rerunning the tests or changing the download url?

maxhaton:
> so what's left here? rerunning the tests or changing the download url?

We updated the runner, but old script doesn't play nice with the new LLVM cmake setup and this version of VS. I think I'm just going to replace it with the instructions from the llvm site. The only potential issue is the SEH thing.

PR #510, opened 2022-06-28
support customizing install directory with $DPATH

WebFreak001:
This introduces the $DPATH environment variable to override the user's home directory just for D tools.

This is also intended to be used by DUB (https://github.com/dlang/dub/pull/2281) when there is not the more explicit `$DUB_HOME` variable set on the system.

This allows users who install their D compiler through the install.sh script to both have the compiler installations as well as the dub packages be stored elsewhere in the system than in their home folder.

In case we have other tools that download their dependencies into the user's home directory too, we should also make them support the same $DPATH environment variable. This way we have a unified and configurable placement location for all the D tools caches and dependencies in the ecosystem which are installed through traditional unix ways.


WebFreak001:
NOTE: script changes untesteed!

PR #513, opened 2022-07-09
Adapt to DMD monorepo with merged druntime

kinke:


kinke:
[Had to move this to my repo: #514.]

PR #523, opened 2022-10-03
[github] Bump FreeBSD VM host image to macOS-12

ibuclaw:
Github is removing 10.15 soon.  Bumped the related actions to their latest versions, to ensure that they work with newer versions of macOS hosts.

PR #524, opened 2022-10-03
Use downloads.dlang.org instead of amazonaws for secondary mirror

ibuclaw:
This is using our new downloads bucket, the old s3 bucket is there for archive only (no longer receiving any updates).


ibuclaw:
Blocked by #523.

PR #527, opened 2022-10-11
Remove workarounds for old OSX release image

ibuclaw:
The old release builds were being built on a 10.8 machine, I've now ported it across to a 10.13 machine.  Still old, but not so much so that we have to build differently to Linux/Windows anymore.

### DIFF
--- a/create_dmd_release/build_all.d
+++ b/create_dmd_release/build_all.d
@@ -30,7 +30,7 @@ enum linux_both = Platform(OS.linux, Model._both);
 /// OSes that require licenses must be setup manually
 
 /// Name: create_dmd_release-osx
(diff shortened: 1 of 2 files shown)


kinke:
AFAIK, env variable `MACOSX_DEPLOYMENT_TARGET=10.12` controls the min OS version the executable will work on. Absolutely no idea what the default setting is if that variable isn't set; LDC CI always sets it explicitly on macOS (10.12 since v1.30, 10.9 for a while before that).

ibuclaw:
@kinke dmd is built with `MACOSX_DEPLOYMENT_TARGET=10.9` - that ldc using 10.12 has not gone unnoticed, I even see warnings spam the CI pipelines too.

kinke (approved):

PR #532, opened 2022-10-18
Actions: Upgrade version of actions used & fix building dub nightly

Geod24:
This should clear up a few deprecations.


Geod24:
@ibuclaw : I fixed dub master but it's not being picked up by the script because the logic around `dub` differs.
Added a commit to fix this, which should hopefully resolve the situation with the nightlies.

ibuclaw (approved):

PR #534, opened 2022-11-30
Build the Linux release on Ubuntu 20.04

Geod24:
Building on ubuntu-latest may result in more recent libraries being linked in. E.g. currently dub nightly build fails to run dub-master because it depends on glibc 2.32+ while ubuntu-20.04 only has 2.31.


Geod24:
image

PR #551, opened 2022-12-11
Update mime types icons

jordisayol:
Add mime types icons for .dd and .di text files


jordisayol:
split mime types into three files types

PR #554, opened 2023-01-05
Nobody has routinely updated Visual D for 3 years

LorenDB:
What can I say?


ibuclaw:
I think this should target stable so it goes in the next release candidate. Also a bug report or changelog entry would be good to give this visibility.

LorenDB:
@ibuclaw: done and done!

rainers:
Thanks. Before VS 2022 an update was not strictly necessary, as Visual D would suggest updating and help with downloading if a newer version is available. But since VS 2022 is now a 64-bit process, older Visual D extensions would not run at all.

rainers (approved):

PR #565, opened 2023-12-17
Adapt to generic Makefiles

kinke:
Depends on
* https://github.com/dlang/dmd/pull/15894 and https://github.com/dlang/dmd/pull/15920
* https://github.com/dlang/phobos/pull/8862
* https://github.com/dlang/tools/pull/465

Soft depends on
* https://github.com/dlang/installer/pull/566 

The Windows VM used for the releases needs a GNU make; `mingw32-make.exe` is available for CI (the first `make.exe` in PATH there is DigitalMars make), so I went with that for now. FWIW, I've just seen that GNU make can also be installed via Chocolatey (`choco install make`).

### DIFF
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -417,21 +417,10 @@ void buildAll(Bits bits, string branch)
         info("Building Dub "~bitsDisplay);
         changeDir(cloneDir~"/dub");
 
-        if (exists("build.d"))
-        {
-            // v1.20+
-            version (Windows)
-                run(msvcVars~"SET DMD="~hostDMD~" && "~hostDMD~" -m"~bitsStr~" -run build.d -O -w -m"~bitsStr);
-            else
-                run("DMD="~hostDMD~" "~hostDMD~" -run build.d -O -w -m"~bitsStr);
-        }
+        version (Windows)
+            run(msvcVars~"set DMD="~hostDMD~"&& "~hostDMD~" -m"~bitsStr~" -run build.d -O -w -m"~bitsStr);
         else
-        {
-            version (Windows)
-                run(msvcVars~"SET DC="~hostDMD~" && build.cmd -m"~bitsStr); // TODO: replace DC with DMD
-            else
-                run("DMD="~hostDMD~" ./build.sh -m"~bitsStr);
-        }
+            run("DMD="~hostDMD~" "~hostDMD~" -run build.d -O -w -m"~bitsStr);
         rename(cloneDir~"/dub/bin/dub"~exe, cloneDir~"/dub/bin/dub"~bitsStr~exe);
     }
 }
--- a/create_dmd_release/create_dmd_release.d
+++ b/create_dmd_release/create_dmd_release.d
@@ -364,13 +364,10 @@ void buildAll(Bits bits, string branch)
     info("Building Druntime "~bitsDisplay);
     changeDir(cloneDir~"/dmd/druntime");
     run(msvcVars~makecmd~pic);
-    removeFiles(cloneDir~"/dmd/druntime", "*{"~obj~"}", SpanMode.depth,
-        file => !file.baseName.startsWith("minit"));
 
     info("Building Phobos "~bitsDisplay);
     changeDir(cloneDir~"/phobos");
     run(msvcVars~makecmd~pic);
-    removeFiles(cloneDir~"/phobos", "*{"~obj~"}", SpanMode.depth);
 
     version(Windows) if (is32)
     {
@@ -379,8 +376,6 @@ void buildAll(Bits bits, string branch)
         info("Building Druntime 32omf");
         changeDir(cloneDir~"/dmd/druntime");
         run(makecmd_omf);
-        removeFiles(cloneDir~"/dmd/druntime", "*{"~obj~"}", SpanMode.depth,
-                    file => !file.baseName.startsWith("minit"));
 
         info("Building OMF import libraries");
         changeDir(cloneDir~"/dmd/druntime/def");
@@ -389,7 +384,6 @@ void buildAll(Bits bits, string branch)
         info("Building Phobos 32omf");
         changeDir(cloneDir~"/phobos");
         run(makecmd_omf);
-        removeFiles(cloneDir~"/phobos", "*{"~obj~"}", SpanMode.depth);
     }
 
     // Build docs
@@ -418,8 +412,6 @@ void buildAll(Bits bits, string branch)
         info("Building Tools "~bitsDisplay);
         changeDir(cloneDir~"/tools");
         run(tools_makecmd~" rdmd ddemangle dustmite");
-
-        removeFiles(cloneDir~"/tools", "*.{"~obj~"}", SpanMode.depth);
     }
 
     bool buildDub = true; // build64BitTools || is32;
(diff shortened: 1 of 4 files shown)


thewilsonator (approved):

ibuclaw:
So, without this patch, master fails to build on windows with:
```
Error: can't read makefile 'win32. mak' 
```

ibuclaw:
With this patch:
```
'mingw-make is not recognised as an internal or external command,
operable program or batch file.
```

ibuclaw:
Searching for anything that looks like a make command inside the windows release box, there's only
- dm/bin/make
- VC/Tools/MSVC/.../bscmake
- VC/Tools/MSVC/.../nmake
- VC/Tools/MSVC/.../xcdmake

ibuclaw:
The mingw project on sourceforge has prebuilt binaries for mingw32-make. Fumbling around seeing if I can build a derivative vagrant box with just that file in the path added.

ibuclaw:
> The mingw project on sourceforge has prebuilt binaries for mingw32-make.

That failed with
```
makefile:32: Extraneous text after `else' directive
makefile:34: Extraneous text after `else' directive
makefile:34: *** only one `else' per conditional.  Stop.
```
The version that worked inside the box was 3.80. Looking at the makefile, I see there's a download link to 4.4 on the dmd github releases page, duh.

ibuclaw:
Alright, looks like installing gnumake-4.4 inside the box is doing the trick.
Only (probably harmless) warning I see is
```
Using make SHELL "C:/Program Files/Git/usr/bin/bash.exe", should be bash.
```

Another I saw when it moves to build dlang/tools is
```
'[' is not recognised as an internal or external command,
operable program or batch file.
```
Doesn't break the build, but it was introduced by https://github.com/dlang/tools/pull/465

However, the last step (Copy Sources) fails on Windows at the following location (new comment thread)....

create_dmd_release/create_dmd_release.d
```diff
@@ -511,7 +440,16 @@ void createRelease(string branch)
...
+    static void ensureIsClean(string repoDir)
+    {
+        const output = runCapture("cd "~quote(repoDir)~" && git status --porcelain");
+        if (output.length)
+            fail("Repo '"~repoDir~"' is dirty:\n" ~ output);
+    }
+
     // Copy sources
+    ensureIsClean(cloneDir~"/dmd");
+    ensureIsClean(cloneDir~"/phobos");
```
ibuclaw, line 452: ... Here.

@schveiguy mentioned this last year, but also again recently in a DLF call. For some reason dmd releases have _always_ had `-dirty` in the version string. No surprises then that this fails.

https://pastebin.com/DfbFk8K4

The list appears to be arbitrary, so I suspect it's some automated unix2dos conversion (git itself is one known culprit for doing this).

ibuclaw: Running the commands individually inside the box, it differs at the first hurdle:
```
old mode 100755
new mode 100644
```
This is every file in the list, so it would appear `scp` command is at fault for not preserving permissions.

ibuclaw: https://github.com/dlang/installer/blob/master/create_dmd_release%2Fbuild_all.d#L153-L159

Instead invoking `scp -r - p` I get either of the following for various files.
```
scp: ... : set times: Input/output error
scp: ... : set times: Broken pipe
```

ibuclaw: https://github.com/dlang/installer/blob/master/create_dmd_release%2Fcreate_dmd_release.d#L297-L308

The best I can think of is to add
```
version (Windows)
    run("git config core.fileMode false");
```
After `git clean` for all repos in that function.

ibuclaw: FAOD, build succeeds with suggested filemode change.

kinke:
Thx for testing/verifying/fixing in the release VM.

> Alright, looks like installing gnumake-4.4 inside the box is doing the trick. Only (probably harmless) warning I see is
> 
> ```
> Using make SHELL "C:/Program Files/Git/usr/bin/bash.exe", should be bash.
> ```

This is no warning, just some info output I've added in https://github.com/dlang/dmd/blob/4712896a98f7d1182a14afaa6fbd83621667bded/compiler/src/osmodel.mak#L69.

> Another I saw when it moves to build dlang/tools is
> 
> ```
> '[' is not recognised as an internal or external command,
> operable program or batch file.
> ```
> 
> Doesn't break the build, but it was introduced by [dlang/tools#465](https://github.com/dlang/tools/pull/465)

This is almost certainly https://github.com/dlang/tools/blob/6670ee433f89a120866e3f0a5a3a551f785e7d0d/Makefile#L13; the problem is that we need DMD's `osmodel.mak`, which also sets up the shell on Windows so that `[.exe` would be found in PATH, but for that we need to make sure we have a DMD clone first, and that's what that line does. For `dlang/tools` CI, this works as-is; `[,exe` is most likely in PATH already (bundled with Git in `C:\Program Files\Git\usr\bin`). - But as long as it's just a non-fatal error during the build, with an existing `..\dmd` clone...

---

What prevents us from moving the release package generation to CI instead of doing it in some AFAICT loosely-defined VM? The existing GHA workflow seems to cover everything; is it just for secrets needed for codesigning on Windows etc.?

ibuclaw:
> > ```
> > '[' is not recognised as an internal or external command,
> > operable program or batch file.
> > ```
> > 
> > Doesn't break the build, but it was introduced by [dlang/tools#465](https://github.com/dlang/tools/pull/465)
> 
> This is almost certainly https://github.com/dlang/tools/blob/6670ee433f89a120866e3f0a5a3a551f785e7d0d/Makefile#L13; the problem is that we need DMD's `osmodel.mak`, which also sets up the shell on Windows so that `[.exe` would be found in PATH, but for that we need to make sure we have a DMD clone first, and that's what that line does. For `dlang/tools` CI, this works as-is; `[.exe` is most likely in PATH already (bundled with Git in `C:\Program Files\Git\usr\bin`). - But as long as it's just a non-fatal error during the build, with an existing `..\dmd` clone...

I think you can use `wildcard` as a platform agnostic (if a little non-intuitive) way to determine if a directory exists in GNU make.


> What prevents us from moving the release package generation to CI instead of doing it in some AFAICT loosely-defined VM? The existing GHA workflow seems to cover everything; is it just for secrets needed for codesigning on Windows etc.?

All build steps are here 

https://gist.github.com/ibuclaw/35d19c5ff64ecea25e4a7c7ce6a79eac#file-release-sh-L53https://gist.github.com/ibuclaw/35d19c5ff64ecea25e4a7c7ce6a79eac#file-release-sh-L53  

Off the top of my head, at mininum there are
- b2/aws secrets for uploading to download dlang.org
- ssh access and permissions to upload to ftp.digitalmars.com
- commit access to dmd, dub, dlang.org, plus ability to push signed tags to these and phobos, tools, and installer
- gpg key for signing tags, packages, tarballs, ...

I don't understand how codesigning is meant to work in an automated context, it all looks 5x more expensive compared to 5-6 years ago, and now it's impossible to do it without a hardware token anyway (or pay even more for cloud signing).

kinke:
> Off the top of my head, at mininum there are
> 
> * b2/aws secrets for uploading to download dlang.org
> * ssh access and permissions to upload to ftp.digitalmars.com
> * commit access to dmd, dub, dlang.org, plus ability to push signed tags to these and phobos, tools, and installer
> * gpg key for signing tags, packages, tarballs, ...
> 
> I don't understand how codesigning is meant to work in an automated context, it all looks 5x more expensive compared to 5-6 years ago, and now it's impossible to do it without a hardware token anyway (or pay even more for cloud signing).

Maybe we can split this process up, e.g., via a dedicated CI runner under our control featuring the required secrets. So that it could trigger the release tagging first, then let regular GHA generate the 'raw' install packages, and then post-process (codesigning etc.) & upload these packages on the extra CI runner again. This way, the extra CI runner would probably have way less prerequisites regarding the software stack, not needing the full build toolchain as the current VM(s), which is duplicated for the existing GHA workflows anyway.

create_dmd_release/create_dmd_release.d
maxhaton: Is the version(Windows) necessary?

ibuclaw: If permissions of a file change when building on a proper filesystem, we want the build to fail.

ibuclaw:
> Maybe we can split this process up, e.g., via a dedicated CI runner under our control featuring the required secrets. So that it could trigger the release tagging first, then let regular GHA generate the 'raw' install packages, and then post-process (codesigning etc.) & upload these packages on the extra CI runner again. This way, the extra CI runner would probably have way less prerequisites regarding the software stack, not needing the full build toolchain as the current VM(s), which is duplicated for the existing GHA workflows anyway.

Could do. I (and @Geod24) have the servers where secrets can be put.

We can create a dlang-release account on github I guess (unless they have the equivalent of a gitlab bot account) - with its own ssh and gpg keys.

I suspect though that it would be simpler to have all repos merged into one though?

ibuclaw:
Merging, otherwise will not be able to build the next beta.

PR #568, opened 2024-01-13
Bump FreeBSD to 13.2

tim-dlang:
The DMD nightly build for FreeBSD fails since two weeks, because the old version is not available anymore. As a result the nightly builds for other platforms are also not uploaded.

This increases the versions for FreeBSD the same as in https://github.com/dlang/dmd/pull/15979, but I have not tested it.


thewilsonator (approved):

PetarKirov:
Thank you, @tim-dlang!

PR #569, opened 2024-01-22
Only use an absolute path in the activate script.

JohanEngelen:
This fixes the activate script when install.sh is called with a relative path (e.g. `install.sh -p .`).


JohanEngelen:
CI failures are not related to this change.

PR #576, opened 2024-05-09
[stable] GHA: Switch from macos-latest (currently *arm64* macOS 14) to macOS 12

kinke:


kinke:
Okay, `macos-12` works; I'll try macos-13 shortly too.

kinke:
Nope, same problem (Mac linker assertion) on macOS 13 x86_64. I guess it's Xcode v15 then, as both macos-{13,14} images currently default to v15.0.1, while macos-12 defaults to v14.2.

ibuclaw:
Ok, I'll merge this as-is for now as I see everything is green. Investigating Xcode 15 can then be done separately.

PR #578, opened 2024-05-25
Windows: Remove obsolete support for OMF/DMC/Optlink toolchain

kinke:


thewilsonator (approved):

kinke:
I've noticed that the changes to the `create_dmd_release/extras` overlay surprisingly aren't applied for the CI artifacts. The problem is that `build_all.d` clones all repos, incl. the `installer` repo itself, and uses that tree for the `extras` (looks like that's the only problematic subtree). For PRs like this from a forked repo, it uses the master branch of `installer` and its extras. It's only after merging that CI uses the new `extras`, and might break. FYI @ibuclaw

As a workaround, one can presumably push the branch to the upstream remote directly.

PR #581, opened 2024-05-27
merge stable

ibuclaw:
- GHA: Switch from macos-latest (currently *arm64* macOS 14) to macOS 12
- GHA: Bump actions/{checkout,{up,down}load-artifact} versions
- Remove all references to optlink
- Windows: Fix missing bin64\sc.ini


ibuclaw:
Well then, with all the recent changes to both branches simultaneously, _something_ appears to have gone wrong @kinke

ibuclaw:
Ahh, the default action I have is "rebase and merge", which is obviously wrong - we want to create a merge commit here instead.

ibuclaw:
No-op for changes -> creating merge commit.

PR #585, opened 2024-06-11
GHA: Replace deprecated set-output

tim-dlang:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This should fix a warning for the nightly builds in bugzilla issue 24600.


kinke (approved):
Thanks!

PR #719, opened 2025-05-19
[stable] Fix CI incl. nightly DMD builds

kinke:
The nightlies are broken since March 6th...


kinke:
Filed https://github.com/dlang/dub/issues/3030 wrt. Validate-results failure on Windows. Now re-targeting `stable`, hopefully that works.

Issue #715, opened 2024-02-13
Installer not working: Can't check signature: No public key


dlangBugzillaToGithub:
### CC List

* Martin Nowak (@MartinNowak)
* Iain Buclaw (@ibuclaw)
* Matthew Cook
* Robert

### Description

````markdown
I cannot install D because the installer is failing right away when trying to verify the install.sh script:

robert@fedora:~$ curl -fsS https://dlang.org/install.sh | bash -s dmd
Downloading https://dlang.org/d-keyring.gpg
######################################################################## 100.0%
Downloading https://dlang.org/install.sh
######################################################################## 100.0%
gpg: Signature made Mon 02 Jan 2023 01:47:04 PM CET
gpg:                using RSA key 30AE2FC45DE4153268ED91754CF5FA5326CC62EB
gpg: Can't check signature: No public key
Invalid signature https://dlang.org/install.sh.sig

I am on Fedora 39 (I know there is an RPM but I still wanted to report this bug).
````

**Comment by dlangBugzillaToGithub** [comment]
#### code (@MartinNowak) commented on 2024-09-25T15:21:35Z



````markdown
More specifically, the keyring was updated by now, but it wasn't signed with any of the existing keys, hence `install.sh update` might fail verifying the new d-keyring. Maybe we should skip that verification step to not complicate bootstrapping that much.

> A workaround is to import the updated key manually for now.
> `gpg --recv-keys E22EC04C82780970381402F4A7D4D42F8EC6A355`

Actually that should update the dlang keyring, not the default one.

`gpg --keyring ~/dlang/d-keyring.gpg --no-default-keyring --recv-keys E22EC04C82780970381402F4A7D4D42F8EC6A355`
````

**Comment by robertschulze** [comment]
This appears to not be working anymore:
```
[root@pve-onedrive ~]# curl -fsS https://dlang.org/install.sh | bash -s dmd
Downloading https://dlang.org/d-keyring.gpg
######################################################################## 100.0%
Downloading https://dlang.org/install.sh
######################################################################## 100.0%
gpg: Signature made Mon Apr  1 20:02:30 2024 UTC
gpg:                using RSA key E22EC04C82780970381402F4A7D4D42F8EC6A355
gpg: Can't check signature: No public key
Invalid signature https://dlang.org/install.sh.sig
[...]

PR #724, opened 2025-07-02
Added a -f,--force option, which disables key signature verification.

everhart:
Currently, the install script is failing because gpg is unable to verify signatures for the dlang-installer, and dmd. This is with fresh gpg keys downloaded by the install script. Seeing as I can't upload up-to-date keys myself, to remedy this I added a "-f,--force" option to the install script. If used, it warns the user signature verification is disabled, and does not perform signature verification. 

I'm not overly concerned about a malicious actor bypassing my computer's security by giving me a malignant D installer, so I would like to have this option as a workaround when the D keyring is out of date. 

This serves as a workaround for Issue #715. I didn't mark it as a fix, as it is merely a workaround, and not a fix.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -125,6 +125,7 @@ download_without_verify() {
     download "$1" 0 "${@:2}"
 }
 
+FORCE=
 # urls...
 fetch() {
     local mirrors path
@@ -390,6 +391,7 @@ parse_args() {
     local _installAction=
     local _getPathAction=
     local _autoInstallFlag=
+    local _force=
 
     while [[ $# -gt 0 ]]; do
         case "$1" in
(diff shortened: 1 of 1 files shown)


CyberShadow:
> I'm not overly concerned about a malicious actor bypassing my computer's security by giving me a malignant D installer, so I would like to have this option as a workaround when the D keyring is out of date.

That doesn't accurately represent the danger - it could be any actor that can make your computer think it's connecting to `dlang.org`. This could be anyone who is capable of MITM (e.g. if they obtain a leaked trusted root certificate private key), or anyone who can access the website (e.g. by compromising the server, stealing credentials, or abusing granted privileges).

`--force` is almost never a good name due to inherent ambiguity, and it's especially not a good name in this situation. Consider instead `--insecure-disable-integrity-check` (with no short alias).

PR #745, opened 2026-07-02
Update installer script to support UCRT and check for the versions of VS that support UCRT.

LightBender:
This PR updates the installer to note the use of UCRT and the VS detection logic to look for all versions of VS that support the UCRT.

This PR is paired with https://github.com/dlang/dmd/pull/23345.

LLM's were used to assist the author of these changes.

### DIFF
--- a/windows/d2-installer-descriptions.nsh
+++ b/windows/d2-installer-descriptions.nsh
@@ -7,7 +7,7 @@ LangString DESC_Dmd2Files ${LANG_ENGLISH} "Digital Mars D version 2 compiler"
 LangString DESC_SMShortcuts ${LANG_ENGLISH} "Add Shortcuts to the Start Menu to start the command interpreter with adjusted environment"
 LangString DESC_AddD2ToPath ${LANG_ENGLISH} "Modify the PATH environment variable so DMD can be used from any command prompt"
 
-LangString DESC_VisualDDownload ${LANG_ENGLISH} "Visual Studio package providing both project management and language services. It works with Visual Studio 2008-2017 (and the free VS Shells)"
+LangString DESC_VisualDDownload ${LANG_ENGLISH} "Visual Studio package providing both project management and language services. It works with Visual Studio 2019 and later"
 
 
 ; Shortcuts
--- a/windows/d2-installer-descriptions.nsh
+++ b/windows/d2-installer-descriptions.nsh
@@ -7,7 +7,7 @@ LangString DESC_Dmd2Files ${LANG_ENGLISH} "Digital Mars D version 2 compiler"
 LangString DESC_SMShortcuts ${LANG_ENGLISH} "Add Shortcuts to the Start Menu to start the command interpreter with adjusted environment"
 LangString DESC_AddD2ToPath ${LANG_ENGLISH} "Modify the PATH environment variable so DMD can be used from any command prompt"
 
-LangString DESC_VisualDDownload ${LANG_ENGLISH} "Visual Studio package providing both project management and language services. It works with Visual Studio 2019 and later"
+LangString DESC_VisualDDownload ${LANG_ENGLISH} "Visual Studio package providing both project management and language services."
 
 
 ; Shortcuts
--- a/windows/d2-installer.nsi
+++ b/windows/d2-installer.nsi
@@ -45,32 +45,28 @@
 ; --------------------
 ; Files
 !define VisualDFilename "VisualD-v${VersionVisualD}.exe"
-!define VS2013Filename "vs_community2013.exe"
-!define VS2017Filename "vs_community2017.exe"
-!define VS2017BTFilename "vs_BuildTools2017.exe"
-!define VS2019Filename "vs_community2019.exe"
-!define VS2019BTFilename "vs_BuildTools2019.exe"
-!define VCRedistx86Filename "vcredist_x86.exe"
-!define VCRedistx64Filename "vcredist_x64.exe"
+!define VS2026Filename "vs_community2026.exe"
+!define VS2026BTFilename "vs_BuildTools2026.exe"
+!define VCRedistx86Filename "vc_redist.x86.exe"
+!define VCRedistx64Filename "vc_redist.x64.exe"
 
 ; URLs
 !define VisualDBaseURL "https://github.com/dlang/visuald/releases/download"
 
 !define VisualDUrl "${VisualDBaseURL}/v${VersionVisualD}/${VisualDFilename}"
 
-!define VS2013Url "http://go.microsoft.com/fwlink/?LinkId=517284"
-!define VS2017Url "https://download.visualstudio.microsoft.com/download/pr/100404311/045b56eb413191d03850ecc425172a7d/vs_Community.exe"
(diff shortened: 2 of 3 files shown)


thewilsonator (approved):

windows/d2-installer-descriptions.nsh
```diff
@@ -7,7 +7,7 @@ LangString DESC_Dmd2Files ${LANG_ENGLISH} "Digital Mars D version 2 compiler"
 LangString DESC_SMShortcuts ${LANG_ENGLISH} "Add Shortcuts to the Start Menu to start the command interpreter with adjusted environment"
 LangString DESC_AddD2ToPath ${LANG_ENGLISH} "Modify the PATH environment variable so DMD can be used from any command prompt"
 
-LangString DESC_VisualDDownload ${LANG_ENGLISH} "Visual Studio package providing both project management and language services. It works with Visual Studio 2008-2017 (and the free VS Shells)"
+LangString DESC_VisualDDownload ${LANG_ENGLISH} "Visual Studio package providing both project management and language services. It works with Visual Studio 2019 and later"
```
rainers: Visual D should still work fine on VS 2012+ (one bug report for VS 2008). If we detect still VS 2015 below, we don't have to be more restrictive here.

windows/d2-installer.nsi
```diff
@@ -45,32 +45,28 @@
 ; --------------------
 ; Files
 !define VisualDFilename "VisualD-v${VersionVisualD}.exe"
-!define VS2013Filename "vs_community2013.exe"
-!define VS2017Filename "vs_community2017.exe"
-!define VS2017BTFilename "vs_BuildTools2017.exe"
-!define VS2019Filename "vs_community2019.exe"
```
rainers, line 51: dmd still comes with a 32-bit executable. VS2019 is the last VS version that runs on a 32-bit system. If we want to support that, we might want to keep that option.

windows/d2-installer.nsi
```diff
@@ -397,6 +394,26 @@ Function DetectVC
         Goto done
     no_vswhere:
 
+    Call DetectVS2026_InstallationFolder
+    StrCpy $1 "VC2026"
+    StrCmp $0 "" not_vc2026 vs2026
+    vs2026:
+        ${LineRead} "$0\VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt" "1" $2
```
rainers, line 428: This won't work on your system that misses this file.

windows/d2-installer.nsi
```diff
@@ -365,19 +358,23 @@ FunctionEnd
 
 Function InstallVCRedistributable
 
-    SetRegView 64 ; look at the 64-bit registry hive if available
+    ; The x86 runtime registers under the 32-bit registry view
+    SetRegView 32
     ClearErrors
-    ReadRegStr $0 HKLM "${VCRedistx86RegKey}" "ProductName"
-    IfErrors 0 vcredistx86_installed
+    ReadRegDWORD $0 HKLM "${VCRedistx86RegKey}" "Installed"
```
rainers, line 391: Is it possible to detect/install the UCRT only, not the VC redistributable? Might avoid running an extra installation if the UCRT is preinstalled with Win11.

rainers (commented):
See also https://github.com/dlang/installer/pull/725 that has become outdated by the release of VS 2026.

windows/d2-installer-descriptions.nsh
LightBender: I did remove the below-2015 detection because UCRT is not supported below 2015, and we are talking about an 11 year old version of VS that is entirely unsupported by Microsoft. I can modify the language to say 2015 and later?

windows/d2-installer.nsi
LightBender: Does DMD need to be called by a 32-bit version of VS? IIRC this should not be a limitation.

windows/d2-installer.nsi
LightBender: Not by itself. UCRT is available on all Windows10/11 installs by default, so we can assume it's presence on those, an OS Major Version number would be sufficient, but below those detection is an ... interesting question. One possible route is to look for the existence of `C:\Windows\System32\ucrtbase.dll`

windows/d2-installer.nsi
LightBender: I can probably make it use the same registry based detection used elsewhere. Let me see about that.

windows/d2-installer-descriptions.nsh
rainers: I'm ok with removing support for the old VS versions. I just meant that the description only mentions support for VS2019 or later, but VS2015 and VS2017 are detected below. Maybe just remove the sentence "It works with...".

windows/d2-installer.nsi
rainers: No, but we install a 32-bit version of dmd to support a 32-bit host-OS. You cannot install VS2022 or VS2026 on that system, though. Either we ignore that, or don't provide the option to install VS on a 32-bit host, or switch to VS2019 in that case.

windows/d2-installer.nsi
rainers: I remember installing the UCRT was quite a hassle on Windows 7. Might be ok to just mention that the VC runtime installer should be used on pre-Win10 systems (does it still work there?) and keep the "do nothing" option.

windows/d2-installer.nsi
rainers: I think you don't need these VS2022/VS2026 detection, DetectVC should be good enough. See also https://github.com/dlang/installer/pull/725/changes#diff-179a67f45668adad5ec8526432aeda299d6c08b4df2bb5349f37489fda1674d4R344

windows/d2-installer.nsi
LightBender: The more recent installation packages of UCRT are much less problematic on pre-Win10 systems. But I have left the Do nothing option available just in case.

windows/d2-installer.nsi
LightBender: DetectVC calls those individual detection methods so we do need to keep them. I am going to leave these as-is for now.

windows/d2-installer.nsi
LightBender: That's a good point I've added an option and then disabled the 2026 installations or the 2019 options based on whether or not the users machine is 64-bit or 32-bit.

windows/d2-installer.nsi
rainers: Ah, ok, detection by vswhere must have already failed when these checks are done. Very unlikely anyway.

windows/vcinstall.ini
```diff
@@ -12,53 +12,45 @@ Bottom=136
...
 [Field 6]
 Type=RadioButton
 Text=Do nothing (you can manually install any of the options above later)
 State=0
 Left=20
 Right=300
-Top=116
-Bottom=130
-
-[Field 7]
```
rainers, line 58: Why was this explanation removed? There is quite a bit of empty space now:

Image

rainers: A usual problem is that the user is required to select the "Desktop development with C++" workload during VS installation. Would be good to mention that here.

LightBender: Looks like it got accidentally deleted while I was moving the Radio buttons around. Repaired.

rainers (approved):

PR #747, opened 2026-09-15
CI: Integrate extra FreeBSD steps in VM

kinke:
Analogous to https://github.com/dlang/dmd/pull/23823, avoiding CI steps duplication (and the need to keep them in sync).

PR #748, opened 2026-09-18
Build compiler with PGO+LTO

kinke:
Except for 32-bit Linux, where PGO isn't possible; only using LTO there.


kinke:
The macOS validation failure is unrelated: https://github.com/dlang/installer/pull/747#issuecomment-5687572338

kinke:
FYI @dkorpel.

kinke:
The current CI problem is also unrelated: https://github.com/dlang/dmd/pull/23886#issuecomment-5795383436

thewilsonator (approved):

dkorpel (approved):
Oh huh, I always assumed pgo was used for release builds.

